Issue #7, opened 2019-07-15
Integrate data-sync layer


cammellos:
As a user
I want my messages to be more reliable 
So that I have more confidence in the app

### Description

This tasks involves integrating https://github.com/status-im/mvds in the actual codebase, according to https://github.com/status-im/bigbrother-specs/blob/master/data_sync/mvds.md#secure-transport.

It should work for one-to-one and private group chats. Pubic chats are out of scope, although a non-breaking path forward should be indicated.

Messages should be authenticated to the original sender.
Messages should be authorized so that only peers that should have access to the messages are able to see/request messages.
Any messages exchanged by the data-sync layer should be PFS encrypted for private communication.

Breaking changes are ok, as long as it's not blocking any possible release pre-v1.

### Acceptance criteria

- MVDS is integrated into the app
- 1-to-1 and group chat messages are exchanged using MVDS
- Messages are authenticated/authorized correctly when relaying them to third party or sending them directly
- A clear non-breaking path forward is indicated for public chats
- This likely will include a change in the actual protocol exchanged on the wire, specs should be updated accordingly before and after implementation: https://github.com/status-im/bigbrother-specs

#### out of scope
- Public chats
- Resource consumption tests

Depends on #2

StatusWrike:
➤ Oskar Thoren commented:

test comment see if wrike syncs

StatusWrike:
➤ Oskar Thoren commented:

Should be synced with https://github.com/status-im/status-protocol-go/issues/7

StatusWrike:
➤ Wrike Bot commented:

Andrea Piana All the necessary task dependencies have been completed, so you can start working on it right away.

cammellos:
> A clear non-breaking path forward is indicated for public chats

Currently datasync layer is between the application and encryption layer, so sending a message over a public channel with datasync is non breaking.

PR #19, opened 2019-07-16
[Part of #7]  Wrap & signs messages if specified in config 

cammellos:
Currently datasync propagates the unencrypted messages. This messages
might be relayed to a third party. We can't therefore rely on whisper
signature, as the author might not be the same as the relayer.
    
This commit wraps our current message in a protobuf record with an
optional signature.
    
If the signature is present it will be used to indicate the author of
the message, otherwise whisper is used.
    
It's not enabled by default as this is a breaking change and should go
live with v1. Decoding is backward compatible, so old messages will
still be read.
    
I have also changed the messageID scheme to use the Keccak256 of the
encoded unencrypted payload, instead of the whisper ID.

## Protocol change
Messages are now wrapped in a protobuf record:
```
bytes payload // Encoded transit
bytes signature //Signature of encoded transit
```

I haven't tested it with the console client yet.

I have passed a bool for the config, but given that there's another PR that does the same we might want to have a config struct, or pass a generic v1 config, although probably having a config struct with separate fields is neater.


messenger.go
```diff
@@ -162,11 +176,27 @@ func NewMessenger(
...
+			datasyncnode.BATCH,
+		)
+		dataSyncNode.Start()
+	}
 
 	messenger = &Messenger{
 		identity:    identity,
 		persistence: &sqlitePersistence{db: messagesDB},
-		adapter:     newWhisperAdapter(identity, t, encryptionProtocol),
+		adapter:     newWhisperAdapter(identity, t, encryptionProtocol, nil),
```
```suggestion
		adapter:     newWhisperAdapter(identity, t, encryptionProtocol, dataSyncNode),
```

PombeirP:
@cammellos I've created a `featureFlags` type in my PR in case you want to leverage that to add `sendV1Messages`.

decanus:
@cammellos one thing we may want to investigate is whether it is worth using https://github.com/gogo/protobuf if we start using protobufs more often. I believe they have a more efficient marshall and unmarshall method.

v1/message.go
```diff
@@ -139,15 +143,50 @@ func CreatePrivateTextMessage(data []byte, lastClock int64, chatID string) Messa
 	return createTextMessage(data, lastClock, chatID, MessageTypePrivate)
 }
 
-// DecodeMessage decodes a raw payload to Message struct.
-func DecodeMessage(data []byte) (message StatusMessage, err error) {
+func unwrapMessage(data []byte) (*StatusProtocolMessage, error) {
+	message := StatusProtocolMessage{}
```
adambabik: Nitpick: using `var message StatusProtocolMessage` you won't allocate memory if `data` is invalid.

v1/message.go
```diff
@@ -139,15 +143,50 @@ func CreatePrivateTextMessage(data []byte, lastClock int64, chatID string) Messa
...
-		return
+		return nil, err
+	}
+	return value, nil
+}
+
+// DecodeMessage decodes a raw payload to StatusMessage struct.
+func DecodeMessage(data []byte) (message StatusMessage, err error) {
+	transitMessage := data
+	message = StatusMessage{}
```
adambabik: Not needed. Named return argument has the same zero value.

v1/message.go
```diff
@@ -139,15 +143,50 @@ func CreatePrivateTextMessage(data []byte, lastClock int64, chatID string) Messa
...
+
+	statusProtocolMessage, err := unwrapMessage(data)
+	if statusProtocolMessage != nil {
```
adambabik: `err` should be checked instead.

v1/message.go
```diff
@@ -139,15 +143,50 @@ func CreatePrivateTextMessage(data []byte, lastClock int64, chatID string) Messa
...
+			if err != nil {
+				return message, err
+			}
+
+			message.SigPubKey = recoveredKey
+		}
 	}
-	return StatusMessage{Message: value}, nil
+
+	value, err := decodeTransitMessage(transitMessage)
```
adambabik: `err` not checked. If we are fine that it fails, we should at least log it.

v1/message.go
```diff
@@ -159,3 +198,25 @@ func EncodeMessage(value Message) ([]byte, error) {
...
 }
+
+func MessageID(data []byte) []byte {
+	return crypto.Keccak256(data)
+}
+
+// WrapMessageV1 wraps a payload into a protobuf message and signs it if an identity is provided
+func WrapMessageV1(payload []byte, identity *ecdsa.PrivateKey) ([]byte, error) {
+	var signature []byte
+	var err error
```
adambabik: Can be declared inside `if` as it's not used outside of it.

v1/message_test.go
```diff
@@ -24,6 +25,7 @@ func TestDecodeMessage(t *testing.T) {
 	require.NoError(t, err)
 	require.EqualValues(t, StatusMessage{
 		Message: testMessageStruct,
+		ID:      crypto.Keccak256(testMessageBytes),
```
adambabik: Let's use `MessageID`.

v1/message_test.go
```diff
@@ -49,6 +51,41 @@ func TestEncodeMessage(t *testing.T) {
 	require.NoError(t, err)
 	require.EqualValues(t, StatusMessage{
 		Message: testMessageStruct,
+		ID:      crypto.Keccak256(data),
```
adambabik: Here as well.

v1/message_test.go
```diff
@@ -49,6 +51,41 @@ func TestEncodeMessage(t *testing.T) {
...
+	key, err := crypto.GenerateKey()
+	require.NoError(t, err)
+
+	data, err := EncodeMessage(testMessageStruct)
+	require.NoError(t, err)
+	wrappedMessage, err := WrapMessageV1(data, key)
+	require.NoError(t, err)
+	// Decode it back to a struct because, for example, map encoding is non-deterministic
+	// and it is not possible to compare bytes.
+	val, err := DecodeMessage(wrappedMessage)
```
adambabik: Hm. Looking at it from lib user perspective, we have separate `EncodeMessage` and `WrapMessageV1` but only `DecodeMessage`. Maybe we should have `EncodeMessage(WrapMessageV1(data))` and its counterpart `DecodeMessage(UnwrapMessageV1(data))`.

oskarth:
@cammellos Can we create a corresponding PR in the https://github.com/status-im/specs repo?

adapters.go
```diff
@@ -30,16 +30,18 @@ const (
 // whisperAdapter is a bridge between encryption and transport
 // layers.
 type whisperAdapter struct {
-	privateKey *ecdsa.PrivateKey
-	transport  transport.WhisperTransport
-	protocol   *encryption.Protocol
+	privateKey     *ecdsa.PrivateKey
+	transport      transport.WhisperTransport
+	protocol       *encryption.Protocol
+	sendV1Messages bool
```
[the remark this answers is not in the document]
oskarth: I find sendV1Messages a bit confusing, can we document it a bit more in the code?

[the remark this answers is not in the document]
cammellos: @PombeirP already implemented flags in his PR , I will port the code over

adapters.go
```diff
@@ -152,8 +154,10 @@ func (a *whisperAdapter) decodeMessage(message *whisper.Message) (*protocol.Stat
 	if err != nil {
 		return nil, err
 	}
-	decoded.ID = message.Hash
-	decoded.SigPubKey = publicKey
+	// Don't override signature if present
```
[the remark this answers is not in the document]
cammellos: what do you mean exactly?

v1/message_test.go
cammellos: I feel a single DecodeMessage should be ok, as we are always fine decoding V1 messages, while sending we need to make a choice (also by the time we go to v1 we won't have `WrapMessageV1` but only `EncodeMessage` and `DecodeMessage`), if you don't feel strongly about it I would leave it as it for now

cammellos:
@adambabik addressed feeback, I have added feature flags from @PombeirP PR. Only one point I am not sure ("Let's add a scenario when it can be already provided."). @oskarth I will open a pr against specs

cammellos:
@oskarth https://github.com/status-im/specs/pull/34

messenger.go
```diff
@@ -41,6 +41,12 @@ type Messenger struct {
 	ownMessages map[string][]*protocol.Message
 }
 
+type featureFlags struct {
+	// sendV1Messages indicates whether we should sent
```
PombeirP:
```suggestion
	// sendV1Messages indicates whether we should send
```

messenger.go
```diff
@@ -41,6 +41,12 @@ type Messenger struct {
...
+	// messages compatible only with V1
```
PombeirP:
```suggestion
	// messages compatible only with V1 and later
```

adapters.go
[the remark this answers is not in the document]
adambabik: In this case, I would add something like this:
```suggestion
	// Getting a signature from the Whisper message should happen only if
	// the signature was not defined in the payload itself.
```

And we can move `publicKey, err := crypto.UnmarshalPubkey(message.Sig)` inside `if decoded.SigPubKey == nil` to avoid unnecessary work in case of wrapped messages.

messenger.go
```diff
@@ -41,6 +41,12 @@ type Messenger struct {
 	ownMessages map[string][]*protocol.Message
 }
 
+type featureFlags struct {
+	// sendV1Messages indicates whether we should send
+	// messages compatible only with V1 and later
```
```suggestion
	// messages compatible only with V1 and later.
	// V1 messages adds additional wrapping
	// which contains a signature and payload.
```

v1/message_test.go
adambabik: Ok, that makes sense. We can clean it up later after V1.

v1/message_test.go
```diff
@@ -49,6 +51,41 @@ func TestEncodeMessage(t *testing.T) {
...
+	require.NoError(t, err)
+	require.EqualValues(t, StatusMessage{
+		Message:   testMessageStruct,
+		ID:        crypto.Keccak256(data),
```
adambabik: Don't we want to use `MessageID`? In case the implementation of it changes, we would need to fix all tests.

adapters.go
```diff
@@ -229,14 +233,26 @@ func (a *whisperAdapter) SendPublic(ctx context.Context, chatName, chatID string
 		return nil, errors.Wrap(err, "failed to encode message")
 	}
 
+	if a.featureFlags.sendV1Messages {
```
adambabik: This is in two places already. Alternatively, we could add `a.encodeMessage(message)` method which would do the wrapping if the flag is set. 

But this is fine as well.

adapters.go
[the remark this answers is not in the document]
cammellos: makes sense

adapters.go
cammellos: good idea

cammellos:
@adambabik @PombeirP could you review again please?

I have changed the ID scheme to be:
`CompressedBytesOfPK` + `Sha3-256(payload)` which is similar to what we do in `status-react`.

We can't have only `Sha3-256(payload)` as potentially an attacker could send us an identical message, resulting in an ID clash.

We therefore use the PK of either transport of provided with the signature.

messenger.go
```diff
@@ -43,6 +43,11 @@ type Messenger struct {
 
 type featureFlags struct {
 	genericDiscoveryTopicEnabled bool
+	// sendV1Messages indicates whether we should send
+	// messages compatible only with V1 and later.
+	// V1 messages adds additional wrapping
```
PombeirP:
```suggestion
	// V1 messages add additional wrapping
```

v1/message.go
```diff
@@ -139,15 +144,59 @@ func CreatePrivateTextMessage(data []byte, lastClock int64, chatID string) Messa
...
+// DecodeMessage decodes a raw payload to StatusMessage struct.
+func DecodeMessage(transportPublicKey *ecdsa.PublicKey, data []byte) (message StatusMessage, err error) {
+	transitMessage := data
+
+	// Getting a signature from transport message should happen only if
+	// the signature was not defined in the payload itself.
+	message.SigPubKey = transportPublicKey
+
+	statusProtocolMessage, err := unwrapMessage(data)
+	// Wrapped message, extract transit and signature
```
PombeirP: Should this comment be inside the `if`?

v1/message_test.go
```diff
@@ -20,35 +21,90 @@ var (
...
+	key, err := crypto.GenerateKey()
+	require.NoError(t, err)
+
+	val, err := DecodeMessage(&key.PublicKey, testMessageBytes)
 	require.NoError(t, err)
 	require.EqualValues(t, StatusMessage{
-		Message: testMessageStruct,
+		Message:   testMessageStruct,
+		SigPubKey: &key.PublicKey,
+		ID:        MessageID(&key.PublicKey, testMessageBytes),
```
PombeirP: Should we test that it fails if we pass a different pubkey (just to ensure that it's indeed taken into account)?

cammellos: Good idea, done, I have added a test for `MessageID`

PR #22, opened 2019-07-16
Add option to enable legacy generic discovery topic support

### LINKED ISSUE #12: Don't listen or publish on the old discovery topic
As a user
I don't want to listen to the discovery topic
So I don't waste too much bandwidth

### Description

For v1 the old discovery topic will not be used anymore, we should keep this behavior for now as we want to still be compatible with pre-v1, but have a toggle (config/flag anything) that disables listening/publishing to this topic. If a client has version 0 of the protocol, we publish instead on the partitioned topic, while for v1 we keep the current behavior (partitioned/negotiated).

### Acceptance criteria

- A flag/config can be set that disable publishing/listening to the old discovery topic

PombeirP:
This PR adds an option to enable legacy generic discovery topic support (disabled by default).

Closes #12 

Notes:
- I noticed that we're not testing e.g. which arguments `WhisperServiceTransport.Init` calls `ChatManager.Init` with. Not sure if that's something we want to do (e.g. with mocks).
- Same thing for `Messenger` and `newWhisperAdapter` (although here we could perhaps test the end result, i.e. that `transport.SendPrivateWithPartitioned` is called when `Messenger` is created with the option disabled).

### DIFF
--- a/adapters.go
+++ b/adapters.go
@@ -33,13 +33,16 @@ type whisperAdapter struct {
 	privateKey *ecdsa.PrivateKey
 	transport  transport.WhisperTransport
 	protocol   *encryption.Protocol
+
+	featureFlags featureFlags
 }
 
-func newWhisperAdapter(pk *ecdsa.PrivateKey, t transport.WhisperTransport, p *encryption.Protocol) *whisperAdapter {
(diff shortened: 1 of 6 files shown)


PombeirP:
What do you think @adambabik?

PR #25, opened 2019-07-17
Structured logging

### LINKED ISSUE #21: Replace log from stdlib with structured logging
As a developer
I want to have fine-grained control over logging
so that I can decide which logs are collected and have better means to analyze them later.

### Description

Currently, all logging is done with `log` package from stdlib. Instead, we should have a performant structured logging package in order to have fine-grained control over the logs.

status-go uses logger from go-etheruem but it's a bit difficult to work with. We can consider implementing an interface so that it's possible to override logger in order to have a single log file after integrating status-protocol-go with status-go.

### Acceptance Criteria

- a single structured logging package is available
- (optional) provide a way to override logger with status-go's one

PombeirP:
Closes #21 

This PR replaces the use of the standard log package with Uber's [zap](https://godoc.org/go.uber.org/zap) structured logging package. The `Messenger` class takes an instance of a logger and propagates it to downstream classes. I selected this library after reading [this article](https://www.client9.com/logging-packages-in-golang/).

I've avoided using the sugared version of the API in order to obtain the absolute maximum performance since the calls to the base version don't look too bad.


adapters.go
```diff
@@ -33,13 +35,18 @@ type whisperAdapter struct {
 	privateKey *ecdsa.PrivateKey
 	transport  transport.WhisperTransport
 	protocol   *encryption.Protocol
+	logger     *zap.Logger
```
adambabik: I am not fan of putting it inside a struct because it is not part of the business logic. Maybe we can just create a global unexported variable? Like
```go
var logger = logger.With(zap.Namespace("whisperAdapter"))
```

Initial `logger` instance can be taken from `log.go`.

PombeirP:
Regarding @adambabik's question about using UTC timestamps, it looks like zap allows you to configure all encoders with https://godoc.org/go.uber.org/zap/zapcore#EncoderConfig
So we could customize `TimeEncoder` with a function that does what we need. That would be done at the caller level though (this PR assumes we receive a pre-configured instance of the zap logger).

adapters.go
```diff
@@ -67,15 +74,19 @@ func (a *whisperAdapter) RetrievePublicMessages(chatID string) ([]*protocol.Mess
...
 	}
 
+	logger := a.logger.With(zap.String("site", "RetrievePublicMessages"))
+
 	decodedMessages := make([]*protocol.Message, 0, len(messages))
 	for _, item := range messages {
 		shhMessage := whisper.ToWhisperMessage(item)
 
-		log.Printf("[whisperAdapter::RetrievePublicMessages] received a public message: %#x", shhMessage.Hash)
+		hlogger := logger.With(zap.String("hash", fmt.Sprintf("%#x", shhMessage.Hash)))
```
adambabik: I wonder how would `zap.Binary()` render this in stdout.

messenger.go
```diff
@@ -80,10 +87,18 @@ func WithChats(
...
+		c.featureFlags.genericDiscoveryTopicEnabled = true
+		return nil
+	}
+}
+
 func NewMessenger(
 	identity *ecdsa.PrivateKey,
 	server transport.Server,
 	shh *whisper.Whisper,
+	logger *zap.Logger,
```
adambabik: Just noticed your comment :) Maybe we can put this into `opts ...Option`? As I mentioned, it's not crucial business logic. Also if not provided, we should default to some standard one. I guess I can pass `nil` now and it will panic.

adapters.go
PombeirP: I thought we'd avoid using global variables (it's also noted in the zap documentation that while supported, they only [advise](https://github.com/uber-go/zap/blob/master/FAQ.md#why-include-package-global-loggers) using the logger as a global variable for legacy apps which can't afford to change APIs). But if that's the way we want to do it, I can move to a global logger.

PombeirP: In a sense, a Mutex is not part of the business logic either, it's an implementation detail.

adapters.go
PombeirP: An example for 
```
zap.Binary("their-identity-key", ecrypto.FromECDSAPub(theirIdentityKey))
```
would be:
```
2019-07-17T17:34:11.508+0200    INFO    encryption/encryptor.go:599     built a message {"encryptor": {"site": "EncryptPayload", "their-identity-key": "BJJcYMg1wdxWVZ82yQMncGdf68fMnrr5TbH8rbx1+Q6Vo55jgzoVeDzEFeULOthfPHkTKsdx/Bc2sxUo139tclk=", "installation-ids": ["1"]}}
```

messenger.go
PombeirP: We could provide a default one, but AFAICT there's no method to shutdown the messanger, where we'd be able to call `logger.Sync()`. Is that something we want to add?

zah:
Not bad
Not bad, not bad.

You may draw some inspiration from how we do structured logging in Nim:
https://github.com/status-im/nim-chronicles/

Or not :) On a more serious note, if you ever go into logging to json, we are developing a tailing tool that might help you as well. It's able to reformat the json output to a more human-readable form and it can do some interactive searching and filtering on the fly.

adapters.go
adambabik: > Avoid them where possible.

It would be great to see some explanation :D I understand that it might be beneficial for testing or when the logger is different for instances.

I can live with this and let's keep it, just don't see any value now.

adambabik: > In a sense, a Mutex is not part of the business logic either, it's an implementation detail.

That's a fair point, however, usually, Mutex is not a part of the business logic. However, in this case, probably the even bigger thing is the fact that `logger` is in the constructor params. If that's unexported constructor, fine. But we need to be careful about exported ones.

messenger.go
```diff
@@ -104,35 +116,50 @@ func NewMessenger(
...
-				log.Printf("[onNewSharedSecretHandler] failed to process secrets: %v", err)
+				slogger := logger.With(zap.String("site", "onNewSharedSecretHandler"))
+				slogger.Warn("failed to process secrets", zap.Error(err))
 			}
-		},
-		onSendContactCodeHandler: func(messageSpec *encryption.ProtocolMessageSpec) {
-			log.Printf("[onSendContactCodeHandler] received a SendContactCode request")
+		}
+	}
+	if c.onSendContactCodeHandler != nil {
```
```suggestion
	if c.onSendContactCodeHandler == nil {
```

messenger.go
```diff
@@ -191,6 +221,12 @@ func NewMessenger(
 	return messenger, nil
 }
 
+// Shutdown takes care of ensuring a clean shutdown of Messenger
+func (m *Messenger) Shutdown() {
+	defer m.logger.Sync()
```
adambabik: `defer` not needed? Just let's put it as the last invocation.

adapters.go
PombeirP: > It would be great to see some explanation :D I understand that it might be beneficial for testing or when the logger is different for instances.

Yeah, easier testing is one of the reasons. Also we'd like to have a different instance for each class because each instance will contain additional context (like the class name it is logging for, and potentially other context in the future. E.g. for testing we're adding the user name, Alice, Bob, etc)

messenger.go
PombeirP: Oops, fixed!

messenger.go
PombeirP: OK, I've added a `Shutdown` method to Messenger with a list of `shutdownTasks` that we register to be run in `Shutdown`. That avoids keeping a pointer to `Logger` in `Messenger`.

PR #26, opened 2019-07-18
Add Messenger.RetrieveAll method

### LINKED ISSUE #24: Provide a way to retrieve messages from all active filters
As a user of the library
I want to be able to retrieve all messages from all active filters
so that I can see messages from chats I haven't created yet.

### Description

Currently, `Messenger` provides `Retrieve()` method but it requires passing a `Chat` instance. 

As we know that on the same Whisper topic there might be many chats, also ones which have been started by another user, we want a way to retrieve all messages for all chats, even not created yet.

This ticket is required for #2, however, #2 has even more requirements like retrieving all types of messages (not only `protocol.Message`) and retrieving raw transit encoded payloads.

This method would be useful for improving status-console-client experience as it does not display chats started by other peers.

### Acceptance criteria

`Messenger.RetrieveAll() (messages []*protocol.Message, err error)` is provided.

PombeirP:
Closes #24

This PR adds a `Messenger.RetrieveAll` method to retrieve all fetched public and private messages.


adambabik:
I started doing something like that for status-go integration: https://github.com/status-im/status-protocol-go/compare/prepare-status-go-integration (I will open a draft PR in a sec).

In the case of #24, I think we can only provide `RetrieveAll` which would match `RetrieveAll() (messages []*protocol.Message, err error)` signature and call it a day. This is needed for status-console-client to support opening new chats not started by us.

PombeirP:
I started by doing a `RetrieveAll` method, but the problem for me is how to distinguish public from private chats in `Messenger` as that is necessary knowledge in order to perform differentiated processing. Right now that knowledge is in `ChatManager` at the Whisper level, and I'm only bubbling up the chat IDs. I'd need to find a better data structure to pass back from the Whisper layer (either an array of `Chat` interfaces, or additional info in `protocol.Message` so that we know if the message is private or not). wdyt?

adambabik:
@PombeirP Here is a case how `Messenger` decides about it in `Send()`: https://github.com/status-im/status-protocol-go/blob/master/messenger.go#L263-L283

PombeirP:
> @PombeirP Here is a case how `Messenger` decides about it in `Send()`: https://github.com/status-im/status-protocol-go/blob/master/messenger.go#L263-L283. Generally, the assumption is that `Chat` for public chats do no return a public key (because there is none).

I've seen that, but in that code you have a `Chat` interface which is passed to you by the consumer of `Messenger`, whereas in this case we don't have that, just a `ChatID` (or whatever we decide to surface up from the Whisper layer, making sure we don't tie ourselves into a specific transport detail).

PombeirP:
Maybe what I could do is to have the Whisper layer only return ChatIDs for public channels (which AFAIU is what @cammellos needs in status-react too)

cammellos:
deciding on public key is not enough, as it does not take into consideration private group chats, so we ll have to change that eventually, and we should not be using that pattern to make decisions.

In this case I would go for something like we had before, we know it works and I see no reason to change it to something that might have some shortcomings (as now that we can't retrieve messages for chats that we don't have open), to minimize unknowns.

I would just enrich `Message` with `ChatID` for public chats,and identity for private channels, we can have a flag/ method `Public/Private` to ease decisions.
 (also those fields should be on all kind of messages, not just `Message`, but that goes for ID, SigPubKey etc), but not 100% sure who and how will use this method.

PombeirP:
@cammellos How usable is `MessageT` for determining the type of message? Is that filled in for received messages?

cammellos:
I think that's not the same, basically `MessageT` is just a inner field of
`Message`, to differentiate the various types of text `Message` (like
system messages, command messages), while I was referring to
`StatusMessage`, which is a wrapper around any type of message, `Message`,
`PairInstallationMessage` etc

@adamb might have more context on this

PombeirP:
So if I understand correctly, I should modify our fork of WhisperV6 to include these fields in `ReceivedMessage`:

```golang

	ChatID   string
	Identity string
```

and adding those fields too to `StatusMessage` (maybe we could avoid adding them to `StatusMessage` if the inner `Message` field implemented an interface with those 2 methods).

adambabik:
`MessageT` is taken from the transit encoding. Currently, it has three values available (this is not verified actually):
```
	MessageTypePublicGroup  = "public-group-user-message"
	MessageTypePrivate      = "user-message"
	MessageTypePrivateGroup = "group-user-message"
```

It is set when creating a message. That's it. 

Regarding the `RetrieveAll`. Like I said, #24 does not care about status-go or status-react. This can just work by going to transport layer and iterating through all filters.

For status-go compliant method, we will return `ChatID` based on the filter like in https://github.com/status-im/status-protocol-go/pull/27.

adapters.go
```diff
@@ -77,32 +77,35 @@ func (a *whisperAdapter) RetrievePublicMessages(chatID string) ([]*protocol.Mess
...
-		case protocol.Message:
-			m.ID = statusMessage.ID
-			m.SigPubKey = statusMessage.SigPubKey
-			decodedMessages = append(decodedMessages, &m)
-		default:
-			hlogger.Error("skipped a public message of unsupported type")
+	var latest []*protocol.ChatMessages
+	for chat, messages := range messagesByChat {
+		decodedMessages, err := a.handleRetrievedMessages(messages, chat.IsPublic())
+		if err == nil { // TODO: What to do when we get an error here? Should it be best effort, or can we safely abort?
```
adambabik: Just log it. We shouldn't stop if there is an invalid message as maybe one can try to kill all clients using a corrupted messages or something.

adapters.go
```diff
@@ -77,32 +77,35 @@ func (a *whisperAdapter) RetrievePublicMessages(chatID string) ([]*protocol.Mess
...
+			if chat.IsPublic() {
```
adambabik: Let's add a comment why we do this only for public chats.

adapters.go
```diff
@@ -155,6 +167,8 @@ func (a *whisperAdapter) RetrievePrivateMessages(publicKey *ecdsa.PublicKey) ([]
 			if err != nil {
 				return nil, err
 			}
+		default:
+			hlogger.Error("skipped a message of unsupported type")
```
```suggestion
			hlogger.Debug("skipped a message of unsupported type")
```

v1/chat_messages.go
```diff
@@ -0,0 +1,7 @@
+package statusproto
+
+type ChatMessages struct {
```
adambabik: I would move it outside of `v1`. In `v1` we want to only store crucial stuff mentioned in the specification.

PombeirP: Wasn't aware of that. I'll add a README.md to the folder with a simple explanation.

adapters.go
PombeirP: The code has changed in the meantime, it is now done for all messages.

PombeirP:
> a few things i am still unclear about is whether we ever need to fetch messages only for a given chat, but that's a minor point and not introduced here.

Isn't that used for the "Fetch last 24 hours" menu item in status-react?

> Another things that jumps out is that we discard our own messages for private chats, which is not currently the correct behaviour, as messages might be sent from a different device, but requires special handling, so might be worth adding a comment, but leave it as it is fir now.

Yeah, in this PR I just focused on reproducing the existing behavior for the other methods.

cammellos:
> Isn't that used for the "Fetch last 24 hours" menu item in status-react?

I don't think is quite the same, fetch 24 hours makes a request to mailserver with a single topic, this one checks a single filter for received messages. 


We never needed the things to be decoupled, although they are in whisper as we always check all filters, because we want to be notified of new messages as soon as they come in, and we always poll all the filters, so we coupled the call (one single rpc call to check all filters), which is much more performing.

PombeirP:
> As i understand the aim of this task is to provide a way for status-console-client to fetch messages for 1 to 1 chats that have not being joined, @adamb please correct me if I am wrong, so I would make sure that the integration works as expected before merging.

@adambabik do you think you could do a quick test in the status-console-client? From what I saw I could consume the API there, but you'd be much more familiar with the right way to integrate the method with the existing code.

adambabik:
> Another things that jumps out is that we discard our own messages for private chats, which is not currently the correct behaviour, as messages might be sent from a different device, but requires special handling, so might be worth adding a comment, but leave it as it is fir now.

We should have a separate ticket in this project to make sure your messages when using device syncing work as expected.

@cammellos it's only in the status-console-client to fetch messages only for a given chat (filter). It looks like a crucial thing to me but of course, from only our perspective and status-react, it's not useful.

@PombeirP the only way is to create a PR in status-console-client, do `go get github.com/status-im/status-protocol-go@hash-of-your-commit` and write integration code in there. When it works, you merge a change in this lib and use  `go get github.com/status-im/status-protocol-go@latest` in the PR in status-console.

adambabik:
I will take this over to integrate it into status-console-client so that new chats appear automatically.

PR #34, opened 2019-07-24
Add datasync

cammellos:
This commit adds the ability of reading and sending datasync messages.
Those messages are PFS encrypted, so the structure of the final messages is:

```
Whisper
--------------
Encryption
--------------
Datasync
---------------
Message 1 Message 2 Message 3
----------------
```

Where Message is either a `protobuf` message wrapping transit, or just `transit` for pre-v1 compatibility.

Datasync messages are batched, so a single whisper message might
contain multiple protocol messages.
Datasync messages are parsed regardless of whether the datasync flag is
on, as this commit is backward compatible.

Sending messages is toggled off by default.

Tested against console client, draft as mvds PR is still pending https://github.com/vacp2p/mvds/pull/61, but feedback is welcome.


adapters.go
```diff
@@ -93,19 +109,26 @@ func (a *whisperAdapter) RetrievePublicMessages(chatID string) ([]*protocol.Mess
 
 		hlogger.Debug("received a public message")
 
-		statusMessage, err := a.decodeMessage(shhMessage)
+		err := a.decryptMessage(context.Background(), shhMessage)
```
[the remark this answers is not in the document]
cammellos: they are not encrypted, but they have bundles attached to it, so probably it's a matter of renaming `decryptMessage` to something like `processEncryptionLayerMessage`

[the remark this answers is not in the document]
cammellos: or something like that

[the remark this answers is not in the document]
cammellos: renamed as `handleEncryptionMessage`

datasync/transport.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (t *DataSyncNodeTransport) AddPacket(p transport.Packet) {
+	t.packets <- p
+}
+
+func (t *DataSyncNodeTransport) Watch() transport.Packet {
+	return <-t.packets
+}
+
+func (t *DataSyncNodeTransport) Send(_ state.PeerID, peer state.PeerID, payload protobuf.Payload) error {
```
oskarth: I don't understand why there are two state.PeerIDs and one is ignored

datasync/utils.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func ToGroupID(data []byte) state.GroupID {
+	g := state.GroupID{}
+	copy(g[:], data[:])
+	return g
+}
+
+// ToOneToOneGroupID returns a groupID for a onetoonechat, which is taken by
+// concatenating the bytes of the compressed keys, in ascending order by X
```
oskarth: Would it make sense to add a link to the relevant part of the spec for this? For future reference

messenger.go
```diff
@@ -61,6 +62,9 @@ type featureFlags struct {
 	// V1 messages adds additional wrapping
 	// which contains a signature and payload.
 	sendV1Messages bool
+
+	// datasync indicates whether messages should be sent using datasync, breaking change for non-v1 clients
```
oskarth: To clarify, this means exclusively datasync? As opposed to in addition. Can we add that in the comment?

Also, how does this impact e.g. public chats in terms of when this flag is set? Not clear from looking at this piece of the code.

datasync/datasync.go
```diff
@@ -0,0 +1,10 @@
+package datasync
+
+import (
+	datasyncnode "github.com/vacp2p/mvds/node"
+)
+
+type DataSync struct {
+	*datasyncnode.Node
+	*DataSyncNodeTransport
```
oskarth: Not sure where this belongs, but could we add a comment on what DataSyncNodeTransport is exactly? Or a link to relevant doc

datasync/transport.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+
+	return t.dispatch(context.TODO(), publicKey, data)
+}
+
+// CalculateSendTime calculates the next epoch
+// at which a message should be sent.
+func CalculateSendTime(count uint64, time int64) int64 {
+	return time + int64(count*2) // @todo this should match that time is increased by whisper periods, aka we only retransmit the first time when a message has expired.
```
oskarth: nice catch for whisper specifics

v1/message.proto
```diff
@@ -3,6 +3,6 @@ syntax = "proto3";
 package statusproto;
 
 message StatusProtocolMessage {
-  bytes signature = 1;
-  bytes payload = 2;
+  bytes signature = 4001;
```
oskarth: Same here as https://github.com/vacp2p/specs/pull/1 would be good to document this more explicitly

cc @adambabik

oskarth:
```
Whisper
--------------
Encryption
--------------
Datasync
---------------
Message 1 Message 2 Message 3
----------------
```

Is there a corresponding PR in https://github.com/status-im/specs repo for this?

cammellos:
@oskarth no, I think though that we should probably keep docs in this repo. First it's easier to keep things in sync, and also the toll currently on making a protocol change can be as high as 6 PRs:

1) Update mvds
2) Update status-protocol-go
3) Update status-go
4) Update status-console-client
5) Update status-react
6) Update specs

So one less PR (specs) would make this less cumbersome (we can still track those file in that repo  as a submodule or whatever, as long as updating code & specs is atomical)

datasync/transport.go
cammellos: that's the interface of `Send` in mvds, the first one is the sender, the other is the destination. If it needs documenting probably the best place is in `mvds` where the interface is defined

oskarth:
Hm, my concern is mainly that the specs repo should be self-contained and allow implementation by another client. If it's just clarification purposes / implementation details then having it as docs in relevant repo seems fine.

I'm not sure why it'd require extra PRs to status-react (outside of normal ones). Having some friction for protocol changes is also not necessarily a bad thing!

@decanus what are your thoughts?

cammellos:
I think we should though take into consideration our workflows.

Who makes changes to the protocol? Status through this repo. 
If at some point that changes, and this repo is not the reference implementation anymore, then we should change it, but for now we would be decoupling to things that are completely coupled.

For visibility, we can still have a repo tracking the docs files, but having them decoupled in our current scenario is providing no benefits and increasing the likelihood of them not being in sync.

In terms of friction, there's definitely enough, any change here, needs to be somehow validated by at least one client, and currently that's really status-react as it's the only fully implemented client (ideally it will be status-console-client at some point), which includes a change in status-go, so that's 5 PR just to make sure the changes are ok. (not validating the changes with a client is also possible, but I don't think it's advisable)

cammellos:
@adambabik ready for review again
@oskarth probably is best if either you or @decanus updates the specs repo, as you are more familiar with the inner working of datasync and we can probably leverage existing docs.
Thanks!

adapters.go
```diff
@@ -134,42 +150,44 @@ func (a *whisperAdapter) handleRetrievedMessages(messages []*whisper.ReceivedMes
 
 		hlogger := logger.With(zap.Binary("hash", shhMessage.Hash))
 
-		hlogger.Debug("handling a received message")
+		hlogger.Debug("received a public message")
```
adambabik: How do we know it's a public message? `handleRetrievedMessages` handles only public messages?

adapters.go
```diff
@@ -212,32 +230,90 @@ func (a *whisperAdapter) RetrieveRaw(filterID string) ([]*whisper.Message, error
...
+}
+
+// decodeMessages expects an unencrypted whisper message as it's argument.
+// 1) Check if it's a datasync message
+// 2) If it's a datasync message it will handle each transmitted message separately
+// 3) if it's not a datasync message a single message will be handled
+// Fingerprinting datasync messages is a bit tricky, as they might unmarshal fine
+// because they are wrapped messages, so we need to check whether anything is
+// poulated in the fields.
+func (a *whisperAdapter) decodeMessages(message *whisper.Message) ([]*protocol.StatusMessage, error) {
```
adambabik: Maybe `handleDecodedMessages()`? We can have a flow like `handleEncryptedMessage()` then `handleDecodedMessages()` and `decodeMessage` would stay a low-level message.

adapters.go
```diff
@@ -212,32 +230,90 @@ func (a *whisperAdapter) RetrieveRaw(filterID string) ([]*whisper.Message, error
...
+			return nil, err
+		}
+
+		decodedMessages = append(decodedMessages, decodedMessage)
+
+	} else {
+		hlogger.Debug("Handling datasync message")
+		// datasync message
+		for _, message := range datasyncMessage.Messages {
+			decodedMessage, err := protocol.DecodeMessage(publicKey, message.Body)
```
```suggestion
			decodedMessage, err := a. decodeMessage(publicKey, message.Body)
```

adambabik: Or get rid of `decodeMessage` but it's misleading to do decoding in two different ways.

adapters.go
```diff
@@ -212,32 +230,90 @@ func (a *whisperAdapter) RetrieveRaw(filterID string) ([]*whisper.Message, error
...
+			a.addDatasyncPacket(publicKey, datasyncMessage)
+		}
+
+	}
+
+	return decodedMessages, nil
 }
 
-func (a *whisperAdapter) decryptMessage(ctx context.Context, message *whisper.Message) error {
+func (a *whisperAdapter) handleEncryptionMessage(ctx context.Context, message *whisper.Message) error {
```
```suggestion
func (a *whisperAdapter) handleEncryptedMessage(ctx context.Context, message *whisper.Message) error {
```
or
```suggestion
func (a *whisperAdapter) handleMessageEncryption(ctx context.Context, message *whisper.Message) error {
```

adapters.go
```diff
@@ -316,30 +392,47 @@ func (a *whisperAdapter) SendPublic(ctx context.Context, chatName, chatID string
 		return nil, errors.Wrap(err, "failed to encode message")
 	}
 
-	newMessage := whisper.NewMessage{
-		TTL:       whisperTTL,
-		Payload:   encodedMessage,
-		PowTarget: whisperPoW,
-		PowTime:   whisperPoWTime,
+	wrappedMessage, err := a.wrapMessageV1(encodedMessage)
```
adambabik: Should we check for `sendV1Messages`?

adapters.go
```diff
@@ -316,30 +392,47 @@ func (a *whisperAdapter) SendPublic(ctx context.Context, chatName, chatID string
...
 }
 
 // SendPublicRaw takes encoded data, encrypts it and sends through the wire.
 // DEPRECATED
 func (a *whisperAdapter) SendPublicRaw(ctx context.Context, chatName string, data []byte) ([]byte, whisper.NewMessage, error) {
-	newMessage := whisper.NewMessage{
+
+	var newMessage whisper.NewMessage
+
+	wrappedMessage, err := a.wrapMessageV1(data)
```
adambabik: Should we respect `sendV1Messages`?

adapters.go
```diff
@@ -397,16 +495,43 @@ func (a *whisperAdapter) SendPrivate(
...
+
+	return protocol.MessageID(&a.privateKey.PublicKey, wrappedMessage), &message, nil
+}
+
+func (a *whisperAdapter) sendWithDataSync(publicKey *ecdsa.PublicKey, message []byte) error {
+	groupID := datasync.ToOneToOneGroupID(&a.privateKey.PublicKey, publicKey)
+	peerID := datasyncpeer.PublicKeyToPeerID(*publicKey)
+	exist, err := a.datasync.IsPeerInGroup(groupID, peerID)
+	if err != nil {
+		return errors.Wrap(err, "peer not in the group")
```
adambabik: Do we know that? Or that there was an error validating this information?

messenger.go
adambabik: What's the relation between this flag and `sendV1Messages`? Do we want to support V1 messages with or without ds?

messenger.go
```diff
@@ -248,6 +277,7 @@ func NewMessenger(
 		shutdownTasks: []func() error{
 			persistence.Close,
 			adapter.transport.Reset,
+			func() error { datasync.Node.Stop(); return nil },
```
adambabik: Should we discard an error in this case?

persistence.go
```diff
@@ -47,7 +47,7 @@ func (db sqlitePersistence) LastMessageClock(chatID string) (int64, error) {
 	}
 
 	var last sql.NullInt64
-	err := db.db.QueryRow("SELECT max(clock) FROM user_messages WHERE contact_id = ?", chatID).Scan(&last)
+	err := db.db.QueryRow("SELECT max(clock) FROM user_messages WHERE chat_id = ?", chatID).Scan(&last)
```
adambabik: Stumbled upon this one as well in #36 :D

adapters.go
cammellos: sorry, wrong log, I botched the rebase

adapters.go
cammellos: `wrapMessageV1` will check the feature flag, it will return the message unchanged or the wrapped payload, the flag will be removed by v1, so I feel checking inside the method means we'll have to remove less code, what do you reckon?

messenger.go
cammellos: the function does not return any error I believe

cammellos:
@adambabik addressed feedback, thanks!

> What's the relation between this flag and sendV1Messages? Do we want to support V1 messages with or without ds?

For now I think it's useful to support both, so we can test the two independently or together (it makes it easier understand where is the bug), and it comes at little complexity cost, before the v1 final cut, we can remove the code as not necessary anymore.

adapters.go
adambabik: Maybe we can call it `tryWrapMessageV1` to indicate that? I agree that it's better to avoid many `if`s but I think we can have here an indicator that it's not definitively a wrapped message but we want to try to wrap it if some conditions are met.

cammellos: sounds good

messenger.go
PombeirP: The PR was merged without the clarification requested, did it fall through the cracks?

cammellos: I could not reply to the thread, no matter how much I refreshed the page, so I had to answer on a comment https://github.com/status-im/status-protocol-go/pull/34#issuecomment-515416417

PombeirP: I was initially referring to Oskar's comment

cammellos: oh I see, yes, forgot to add that

cammellos: I will add in a separate pr

PombeirP: Thanks!

PR #37, opened 2019-07-27
Handle messages in raw method as well

cammellos:
Refactor handling of messages, the logic was getting a bit unwieldy with adding the datasync layer.
The layers are:

`Transport` -> Whisper, provides authentication,integrity
`Encryption` -> PFS, Bundle, multidevice processing, can provide authentication
`DataSync` -> multiplexes messages
`ApplicationMetadata` -> Provides authentication,integrity, canonical messageID
`Application` -> Provides a parsed message

The idea is that any handling is done through this structure

```
type StatusMessage struct {
// TransportMessage is the parsed message received from the trasport layer, i.e the input
TransportMessage interface{}
// ParsedMessage is the parsed message by the application layer, i.e the output
ParsedMessage interface{}

// TransportLayerInfo contains information about the processing of the trasport layer message
TransportLayerInfo LayerInfo
// EncryptionLayerInfo contains information about the processing of the encryption layer message
EncryptionLayerInfo LayerInfo
// DataSyncLayerInfo contains information about the processing of the data-sync layer message
DataSyncLayerInfo LayerInfo
// ApplicationMetadataLayerInfo contains information about the processing of the application-metadata layer message
ApplicationMetadataLayerInfo LayerInfo
}
```

Each layer will be passed this structure in, use the method `Payload() []byte` to get the payload from the previous layer and will set a corresponding `LayerInfo`:

```
type LayerInfo struct {
// The ID calculated at this layer
ID []byte
// SigPubKey is the signature extracted at this layer
SigPubKey *ecdsa.PublicKey
// Error is any non-fatal error that happened at this layer
Error error
// Payload is the payload extracted at this layer, to be passed to the next
Payload []byte
}
```

and return one or (more in some cases, such as `DataSync`) `StatusMessage`.
If an error is returned that means that the pipeline is to be interrupted (say a corrupted signature), otherwise for "soft" errors (say one message is corrupted, while the rest are fine), the `Error` field is set, and the struct is to be passed to the next layer.

The input is an empty struct with `TransportMessage` set.
The output is a fully filled struct with `ParsedMessage` set.

For `Raw` methods we pass a flag to stop processing before `ParsedMessage` is se (i.e we skip the applicationlayer).

I haven't made it more generic that it ought to be as currently there's no need to swap layers, but once that need arise we can make it more generic (interfaces for each layer, same signature etc).


cammellos:
@adambabik what do you think? I haven't gone around fixing all the stuff/test etc, but it would be good to check if you think it's ok as a structure. (there's still some stuff that needs to be moved etc, but overall architecture)

adambabik:
> Refactor handling of messages, the logic was getting a bit unwieldy with adding the datasync layer.

Agree with this :)

Generally, I think it's a good direction. I am not sure about these `TransportMessage` and `ParsedMessage` as interfaces. Also, is that enough to have a single `LayerInfo` type? For example, the encryption layer might provide details for the transport layer when there is a negotiated secret and sending messages.

Another thing is whether we should group functionality into layers. The problem I see is that the state of `StatusMessage` is managed by some other entities instead of the `StatusMessage` itself. Maybe layers can only be a property of the application and message should have behaviours implemented using interfaces?

I also think this adapter thing is a mistake and not needed.

```go
type StatusMessage struct {
    id []byte

    message interface{} // Message, PairMessage, etc.
    encodedMessage []byte // support raw (transit-encoded) messages

    transportMeta struct {
        hash []byte
        sigPublicKey *ecdsa.PublicKey
        chatID string
        public bool
    }
    encryptedPayload []byte   
    
    encryptionMeta struct {
        spec *ProtocolMessageSpec
    }
    decryptedPayload []byte
}

func (m *StatusMessage) Decrypt(dec Decryptor) (err error) {
    // If StatusMessage does not support data sync,
    // m.decryptedPayload would be assigned to m.encodedMessage as well. 
    // It can a feature flag of StatusMessage or a separate struct using composition.
    m.decryptedPayload, err = dec.Decrypt(m.encryptedPayload)
    return
}

func (m *StatusMessage) Demux(datasync DataSync) ([]*StatusMessage, error) {
    messages, err := datasync.Handle(m.decryptedPayload)
    try(err)

    var result []*StatusMessage
    for _, message := range messages {
        sm := m.clone()
        sm.encodedMessage = message
    }

    return result, nil
}

func (m *StatusMessage) Decode(dec Decoder) (err error) {
    m.message, err = dec.Decode(m.encodedMessage)
    return
}
```

In the application, we could have very simple flow:

```go
// Here, we could have methods like handleTransportLayer, handlerEncryptionLayer etc
// if it would make more sense. Especially, if one layer requires a few operations which is 
// not our case for now.
func (app) HandleMessage(m *StatusMessage) ([]*protocol.StatusMessage, error) {
    err := app.tryDecrypt(m)
    try(err)

    messages, err = app.tryDemux(m)
    try(err)

    for _, m := range messages {
        err := app.tryDecode(m)
        try(err)
    }

    return messages, nil
}

func (app) tryDecrypt(m Decryptable) error {
    if app.skipDecrypt { 
        return nil 
    }
    return m.Decrypt(app.decryptor)
}

func (app) tryDemux(m *StatusMessage) ([]*StatusMessage, error) {
    if app.featureFlags.datasync {
        demuxable, ok := m.(Demuxable)
        if !ok { ... }
        return m.Demux(app.datasync)
    }
    return []*StatusMessage{m}
}
```

Sending messages would look similar:

```go
func (app) SendMessage(cfg MessageConfig) ([]byte, error) {
    msg, err := cfg.Build()
    try(err)
    
    err = app.tryEncode(msg)
    try(err)
    
    err = app.tryWrapMessageV1(msg)
    try(err)

    err = app.tryMux(msg)
    try(err)

    err = msg.Encrypt(app.encryptor)
    try(err)

    return app.Send(msg)
}
```

I left the transport layer out of `StatusMessage` in order to simplify things but it could be included as `Receive(Receiver)` method on `StatusMessage`. I included `transportMeta` field which has important transport-layer metadata required for by other layers.

Some fields like `encryptionMeta` would be used only when sending a message. For more clarity, we could have `StatusMessageOut` and `StatusMessageIn`.

cammellos:
@adambabik thanks for the feedback, I have incorporated most of the suggestions, (if not all :) ), I have kept the naming as `HandleXLayer` as in some cases it does more than Decrypting/Demuxing etc, so it might be slightly misleading.
Also for now I have only refactored the receiving flow, I will leave sending for the next PR.

applicationmetadata/service.go
```diff
@@ -0,0 +1,17 @@
+package applicationmetadata
+
+import (
+	"github.com/golang/protobuf/proto"
+)
+
+//go:generate protoc --go_out=. ./message.proto
+
+func Handle(payload []byte) (*Message, error) {
```
adambabik: Too many `handle` methods we have :D 

```suggestion
func Unmarshal(payload []byte) (*Message, error) {
```

datasync/datasync.go
```diff
@@ -1,11 +1,67 @@
...
+
+func (d *DataSync) Add(publicKey *ecdsa.PublicKey, datasyncMessage datasyncproto.Payload) {
+	packet := datasynctransport.Packet{
+		Sender:  datasyncpeer.PublicKeyToPeerID(*publicKey),
+		Payload: datasyncMessage,
+	}
+	d.DataSyncNodeTransport.AddPacket(packet)
+}
+
+func (d *DataSync) Handle(sender *ecdsa.PublicKey, payload []byte) [][]byte {
```
adambabik: I am not sure we can treat datasync as a multiplexer from the caller perspective but if so maybe:
```suggestion
func (d *DataSync) Demux(sender *ecdsa.PublicKey, payload []byte) [][]byte {
```

encryption/migrations/migrations.go
```diff
@@ -1,29 +1,29 @@
-// Code generated by go-bindata. DO NOT EDIT.
+// Code generated by go-bindata.
```
adambabik: We need to use the same tools to generate it. It looks like it's not the case now.

Please check `make install-dev`. It might need updating because `protoc-gen-go@1.3.1` is used while I saw we started using 1.3.2 in some other lib.

v1/status_message.go
```diff
@@ -0,0 +1,147 @@
...
+		return s.ApplicationMetadataLayerSigPubKey
+	}
+
+	return s.TransportLayerSigPubKey
+}
+
+func (s *StatusMessage) Clone() (*StatusMessage, error) {
+	copy := &StatusMessage{}
+
+	err := copier.Copy(&copy, s)
```
adambabik: Have you tried to just copy the struct? Not sure if this is needed. Pretty much all fields are read only and not modified after initialization.

v1/status_message.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	payload, err := enc.HandleMessage(
+		myKey,
+		senderKey,
+		&protocolMessage,
+		m.Hash,
+	)
+
+	// For any error, we return it and assign
```
adambabik:
```suggestion
```

v1/status_message.go
```diff
@@ -0,0 +1,147 @@
...
+	return nil
+
+}
+
+func (m *StatusMessage) HandleApplication() error {
+	originalPayload := m.DecryptedPayload
+	payload := make([]byte, len(originalPayload))
+	copy(payload, originalPayload)
+	// This modifies the payload
+	value, err := decodeTransitMessage(payload)
```
adambabik: How does this function modifies payload? We need to fix this if that's the case.

cammellos: `bytes.NewBuffer` modifies the the `[]byte` array passed in, I can copy it inside `decodeTransitMessage` so that `decodeTransitMessage` does not modify it, what do you think?

datasync/datasync.go
cammellos: I would be a bit cautious about it, as it not only demux, there are also some side-effects involveds (the message might be replicated etc), so I feel `Demux` might be a bit misleading

v1/status_message.go
adambabik:
```
NewBuffer creates and initializes a new Buffer using buf as its initial contents. The new Buffer takes ownership of buf, and the caller should not use buf after this call.
```

Right. Maybe we can do copying in `decodeTransitMessage`? I think it should be clear that whatever payload you pass to `decodeTransitMessage`, it should be unchanged.

v1/status_message.go
cammellos: what do you mean exactly by just copying the struct? i.e "manually", or there some othere method?

cammellos: I see what you mean, not sure about it, it's a bit dangerous as the client probably does not expect that (status-go for example would have to know that it makes a shallow copy, as currently it overrides `Payload` of a whisper message, which is a pointer, with the `DecryptedPayload`), and it's a bit difficult to reason about it (those messages that are demuxed are shallow copies), I would probably delay the decision on this one, until the interfaces are a bit more finalized and err on the side of caution (performance wise should not be a problem as I think we rarely batch datasync messages), what do you think?

PombeirP: Yeah, we need to be very careful in whatever approach we use so as to not make it easy to screw up later by modifying a shared buffer or something like that unintentionally.

adapters.go
```diff
@@ -194,136 +186,80 @@ func (a *whisperAdapter) handleRetrievedMessages(messages []*whisper.ReceivedMes
...
+// layer message, or in case of Raw methods, the processing stops at the layer
+// before
+func (a *whisperAdapter) handleMessages(shhMessage *whisper.Message, applicationLayer bool) ([]*protocol.StatusMessage, error) {
 	logger := a.logger.With(zap.String("site", "decodeMessages"))
-	hlogger := logger.With(zap.Binary("hash", message.Hash))
+	hlogger := logger.With(zap.Binary("hash", shhMessage.Hash))
+	var statusMessage protocol.StatusMessage
 
-	publicKey, err := crypto.UnmarshalPubkey(message.Sig)
+	err := statusMessage.HandleTransport(shhMessage)
```
PombeirP: Would it make sense making this a `NewFromTransport` function instead?

adapters.go
```diff
@@ -194,136 +186,80 @@ func (a *whisperAdapter) handleRetrievedMessages(messages []*whisper.ReceivedMes
...
+
+	}
 
-	// If it failed to decode is not a protobuf message, if it successfully decoded but body is empty, is likedly a protobuf wrapped message
-	if err != nil || !datasyncMessage.IsValid() {
-		hlogger.Debug("Handling non-datasync message")
-		// Not a datasync message
-		decodedMessage, err := protocol.DecodeMessage(publicKey, message.Payload)
+	for _, statusMessage := range statusMessages {
+		err = statusMessage.HandleApplicationMetadata()
```
```suggestion
		err := statusMessage.HandleApplicationMetadata()
```

adapters.go
```diff
@@ -194,136 +186,80 @@ func (a *whisperAdapter) handleRetrievedMessages(messages []*whisper.ReceivedMes
...
-	publicKey, err := crypto.UnmarshalPubkey(message.Sig)
+	err := statusMessage.HandleTransport(shhMessage)
 	if err != nil {
-		return nil, err
+		hlogger.Error("failed to handle transport layer message", zap.Error(err))
 	}
 
-	// Try first with datasync
-	datasyncMessage, err := protocol.UnwrapDatasync(message.Payload)
+	err = a.handleEncryptionLayer(context.Background(), &statusMessage)
```
adambabik: We have a bit of mix of `handle*` messages on the adapter and `Handle*` on the `StatusMessage`. 

To clean up this flow, we could have `a.handle*()` for each layer which would also took care of logging errors.

adapters.go
```diff
@@ -194,136 +186,80 @@ func (a *whisperAdapter) handleRetrievedMessages(messages []*whisper.ReceivedMes
...
 	logger := a.logger.With(zap.String("site", "decodeMessages"))
-	hlogger := logger.With(zap.Binary("hash", message.Hash))
+	hlogger := logger.With(zap.Binary("hash", shhMessage.Hash))
+	var statusMessage protocol.StatusMessage
 
-	publicKey, err := crypto.UnmarshalPubkey(message.Sig)
+	err := statusMessage.HandleTransport(shhMessage)
 	if err != nil {
-		return nil, err
+		hlogger.Error("failed to handle transport layer message", zap.Error(err))
```
adambabik: It looks like we continue for each layer if an error is returned during message processing. Is that ok? Previously, for example, we stopped if we failed to get a signatureo of a message from the transport layer.

adapters.go
cammellos: yes, sounds clearer

v1/status_message.go
adambabik: We can keep deep copy but generally if a value is received and it has slices, they should be copied by the caller if it wants to modify the underlying array. If we don't have this assumption, we would end up in returning deep copies of slices pretty often.

v1/status_message.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (s *StatusMessage) Clone() (*StatusMessage, error) {
+	copy := &StatusMessage{}
+
+	err := copier.Copy(&copy, s)
+	return copy, err
+}
+
+func (m *StatusMessage) HandleTransport(shhMessage *whisper.Message) error {
```
adambabik: I wonder if making `m.Handle*` methods private and moving `StatusMessage` to the root of the project wouldn't make more sense. `StatusMessage` is not really something described in the spec but an implementation detail.

PombeirP:
Great refactor @cammellos, thanks!

v1/message_test.go
```diff
@@ -20,92 +20,31 @@ var (
...
-	val, err := DecodeMessage(&key.PublicKey, data)
-	require.NoError(t, err)
-	require.EqualValues(t, StatusMessage{
-		Message:   testMessageStruct,
-		SigPubKey: &key.PublicKey,
-		ID:        MessageID(&key.PublicKey, data),
-	}, val)
-}
-
-func TestWrappedMessageWithSignature(t *testing.T) {
```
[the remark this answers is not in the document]
cammellos: They are already covered in `adapters_test.go` , but yes I can have more granular tests for the specific methods

adapters.go
cammellos: actually you are right, transport should short circuit, I think the rest are ok to fall through

v1/status_message.go
cammellos: I might do this in a separate PR, once we refactor the sending flow as well, so we have a bit more context

PR #39, opened 2019-07-29
Update datasync comment

cammellos:
Expand datasync explanation and add test for encrypted messages.


cammellos:
thanks for noticing that!

PR #41, opened 2019-07-30
Add chats API calls

cammellos:
This commits adds two API calls:

SaveChat
GetChats

The first one behaves like an upsert.
The second takes two parameters, from, to and returns chats in reverse
order by timestamp.

Members and MembershipUpdates have been stored as blob of data to
simplify the querying (we never query them separately).

I have changed the `Chat` type from interface to a concrete type, the most relevant change is  that `ID` is now a string , for one-to-one it should be set to the hexencoded public key of the user, the chat-name for public chats, and uuid+pk-of-creator for private group chats, in order to be consistent with status-react. 
We append the type in the database to avoid having cases where a public-chat name is the same as a public pk for example.

I will test this against status-react/status-console-client before merging, so please don't merge, but it would be useful to get some feedback, thanks!


chat.go
```diff
@@ -2,8 +2,81 @@ package statusproto
...
-	PublicKey() *ecdsa.PublicKey
+type ChatPagination struct {
+	From uint
+	To   uint
+}
+
+type ChatType int
+
+const (
+	ChatTypeOneToOne = iota
```
adambabik:
```suggestion
	ChatTypeOneToOne = iota + 1
```

to avoid confusion of `ChatType` zero-value and `ChatTypeOneToOne`.

Unless status-react requires it to be `0`.

chat.go
```diff
@@ -2,8 +2,81 @@ package statusproto
...
+	ChatTypePublic
+	ChatTypePrivateGroupChat
+)
+
+type Chat struct {
+	// ID is the id of the chat, for public chats is the name i.e status, for one-to-one
```
adambabik:
```suggestion
	// ID is the id of the chat, for public chats it is the name e.g. status, for one-to-one
```

messenger.go
```diff
@@ -366,36 +366,44 @@ func (m *Messenger) Mailservers() ([]string, error) {
...
+		return m.adapter.LeavePublic(chat.Name)
 	}
 	return errors.New("chat is neither public nor private")
 }
 
+func (m *Messenger) SaveChat(chat Chat) error {
+	return m.persistence.SaveChat(chat)
+}
+
+func (m *Messenger) GetChats(from, to int) ([]*Chat, error) {
```
```suggestion
func (m *Messenger) Chats(from, to int) ([]*Chat, error) {
```

messenger_test.go
```diff
@@ -150,6 +130,250 @@ func (s *MessengerSuite) TestRetrievePrivate() {
...
+	}
+
+	s.Require().NoError(s.m.SaveChat(chat))
+	savedChats, err := s.m.GetChats(0, 10)
+	s.Require().NoError(err)
+	s.Require().Equal(1, len(savedChats))
+
+	actualChat := savedChats[0]
+
+	s.Require().Equal(chat.ID, actualChat.ID)
```
adambabik:
```go
s.Equal(chat, savedChats[0])
```

should have the same effect or this if we have some additional fields not present or having different expected values:
```go
expectedChat := chat
expectedChat.SomeField = SomeValue
s.Equal(expectedChat, savedChats[0])
```

migrations/sqlite/000002_add_chats.up.db.sql
```diff
@@ -0,0 +1,21 @@
...
+unviewed_message_count INT NOT NULL DEFAULT 0,
+last_clock_value INT NOT NULL DEFAULT 0,
+last_message_content_type VARCHAR,
+last_message_content VARCHAR,
+members BLOB,
+membership_updates BLOB
+);
+
+
+DROP TABLE membership_updates;
```
adambabik: Leftover?

persistence.go
```diff
@@ -92,6 +97,154 @@ VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`)
...
+		encodedMembershipUpdates.Bytes(),
+	)
+	if err != nil {
+		return err
+	}
+
+	return err
+}
+
+func (db sqlitePersistence) GetChats(from, to int) ([]*Chat, error) {
```
adambabik:
```suggestion
func (db sqlitePersistence) Chats(from, to int) ([]*Chat, error) {
```

persistence.go
```diff
@@ -92,6 +97,154 @@ VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`)
...
+	// Encode membership updates
+	var encodedMembershipUpdates bytes.Buffer
+	membershipUpdatesEncoder := gob.NewEncoder(&encodedMembershipUpdates)
+
+	if err := membershipUpdatesEncoder.Encode(chat.MembershipUpdates); err != nil {
+		return err
+	}
+
+	// Insert record
+	stmt, err := db.db.Prepare(`INSERT INTO chats(id, name, color, active, type, timestamp, updated_at,  deleted_at_clock_value, public_key, unviewed_message_count, last_clock_value, last_message_content_type, last_message_content, members, membership_updates)
```
adambabik: I would try to expand this similarly to what you did in `GetChats`.

migrations/sqlite/000002_add_chats.up.db.sql
cammellos: we created them in migration v1, but not needed anymore, so I thought we would just drop them (not much of an issue as data migrations are not needed anymore)

chat.go
```diff
@@ -2,8 +2,81 @@ package statusproto
...
+	Signature string `json:"signature"`
+	// Hex encoded public key of the creator of the event
+	From string `json:"from"`
+	// Target of the event for single-target events
+	Member string `json:"member"`
+	// Target of the event for multi-target events
+	Members []string `json:"members"`
+}
+
+// ChatMember represent a member who participates in a group chat
```
PombeirP:
```suggestion
// ChatMember represents a member who participates in a group chat
```

chat.go
```diff
@@ -2,8 +2,81 @@ package statusproto
...
+	Members []ChatMember `json:"members"`
+	// MembershipUpdates is all the membership events in the chat
+	MembershipUpdates []ChatMembershipUpdate `json:"membershipUpdates"`
+}
+
+// ChatMembershipUpdate represent an event on membership of the chat
+type ChatMembershipUpdate struct {
+	// Unique identifier for the event
+	ID string `json:"id"`
+	// Type indicates what kind of event is (i.e changed-name, added-member etc)
```
PombeirP:
```suggestion
	// Type indicates the kind of event (i.e changed-name, added-member etc)
```

chat.go
```diff
@@ -2,8 +2,81 @@ package statusproto
...
+	Type string `json:"type"`
+	// Name represent the name in the event of changing name events
```
PombeirP:
```suggestion
	// Name represents the name in the event of changing name events
```

cammellos:
@adambabik @PombeirP ready to be reviewed again, thanks for the feedback!

PR #46, opened 2019-08-04
Merge filter package into transport/whisper

adambabik:
There is no gain in having a separate `filter` package in `transport/whisper`. This change merges it into `transport/whisper` package.

### DIFF
--- a/transport/whisper/filter.go
+++ b/transport/whisper/filter.go
@@ -1,4 +1,4 @@
-package filter
+package whisper
 
 import (
 	"crypto/ecdsa"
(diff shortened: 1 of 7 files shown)


transport/whisper/filter.go
```diff
@@ -58,25 +58,25 @@ type Chat struct {
...
 	return !c.OneToOne
 }
 
-type ChatsManager struct {
+type filtersManager struct {
 	whisper     *whisper.Whisper
 	persistence *sqlitePersistence
 	privateKey  *ecdsa.PrivateKey
-	keys        map[string][]byte // a cache of symmetric keys derived from passwords
+	keys        map[string][]byte // a cache of symmetric manager derived from passwords
```
PombeirP: @adambabik: what is a "symmetric manager"?

adambabik: Good catch. My editor must have changed that when using "Refactor" feature. I will change this back to `keys`.

PR #48, opened 2019-08-06
Add confirmations for datasync messages.

cammellos:
The signature of the `Raw` methods has been changed so that they return
the application messageID.
Clients will keep track of this ID and receive a notification through
the envelopesMonitor of the state of the envelopes (sent/expired).
Envelopes monitor is passed this ID as well.
We use an array of IDs as datasync messages might be batched.


transport/whisper/envelopes.go
```diff
@@ -0,0 +1,324 @@
+package whisper
+
+import (
+	"context"
+	"errors"
+	"hash/fnv"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
```
adambabik: Let's get rid of this and user `zap.Logger`.

transport/whisper/envelopes.go
```diff
@@ -0,0 +1,324 @@
...
+	m.envelopes[event.Hash] = EnvelopeSent
+	if m.handler != nil {
+		m.handler.EnvelopeSent(m.identifiers[event.Hash], messageID(m.messages[event.Hash]))
+	}
+}
+
+// clearMessageState removes all message and envelope state.
+// not thread-safe, should be protected on a higher level.
+func (m *EnvelopesMonitor) clearMessageState(envelopeID common.Hash) {
+	delete(m.envelopes, envelopeID)
```
adambabik: Tricky mutex management ;) We can fix that later but generally there is no pattern which method assumes a mutex is already acquired when called and which does it by itself.

transport/whisper/whisper_service.go
```diff
@@ -458,236 +386,3 @@ type RetryConfig struct {
 	StepTimeout time.Duration
 	MaxRetries  int
 }
-
-func (a *WhisperServiceTransport) requestMessagesWithRetry(conf RetryConfig, r MessagesRequest) (MessagesResponse, error) {
```
adambabik: Why do we get rid of this? Console Client relies on these methods.

transport/whisper/envelopes.go
```diff
@@ -0,0 +1,324 @@
...
+
+// EnvelopeEventsHandler used for two different event types.
+type EnvelopeEventsHandler interface {
+	EnvelopeSent([][]byte, common.Hash)
+	EnvelopeExpired(common.Hash, error)
+	MailServerRequestCompleted(common.Hash, common.Hash, []byte, error)
+	MailServerRequestExpired(common.Hash)
+}
+
+func messageID(message whisper.NewMessage) common.Hash {
```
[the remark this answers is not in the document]
cammellos: I think we can get rid of this, yes

transport/whisper/envelopes.go
cammellos: this is all copied from status-go, haven't looked at the code much to be honest, probably best to fix later as I would have to familiarize myself with the logic

transport/whisper/whisper_service.go
cammellos: not sure, I believe this might be the old `history` stuff that has been removed from status-console-client (we need to fetch the correct topics), it seems to be compiling ok with this version, do you want to keep it even if it's not currently used?

adambabik: Oh sorry, I forgot. Mailserver support was removed at some point when we started moving stuff around.

cammellos:
@adambabik feedback addressed, ready to be reviewed again, thanks!

PR #49, opened 2019-08-16
Add replies, block contact, remove unused methods

cammellos:
This commit does a few things:
1) Add replies to messages, which are calculated at query time using a
join
2) Add a BlockContact api call that takes care of deleting messages from
a user, deleting the chat, and re-compute the denormalized values. As it
needs to run in a transaction, the method that it relies on take a
transaction as an optional parameter.
3) Change ID scheme for chats/contacts so that it's consistent across
messages/contacts

This is not compatible as I have modified a migration, but it's not a
problem as data compatibility can be broken


persistence.go
```diff
@@ -163,17 +175,16 @@ func (db sqlitePersistence) Chats(from, to int) ([]*Chat, error) {
...
 			&encodedMembers,
 			&encodedMembershipUpdates,
 		)
 		if err != nil {
 			return nil, err
 		}
-
-		// Restore the backward compatible ID
-		chat.ID = chat.ID[:len(chat.ID)-2]
+		chat.LastMessageContent = lastMessageContent.String
```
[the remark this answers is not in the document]
cammellos: It would throw an error,  as strings are not `Nullable`, that's what I saw by using the api, we never had this problem as we always set it to empty string before, but since the field is nullable we should cater for this case

v1/message.go
```diff
@@ -146,7 +147,7 @@ func EncodeMessage(value Message) ([]byte, error) {
 
 // MessageID calculates the messageID from author's compressed public key
 // and not encrypted but encoded payload.
-func MessageID(author *ecdsa.PublicKey, data []byte) []byte {
+func MessageID(author *ecdsa.PublicKey, data []byte) hexutil.Bytes {
```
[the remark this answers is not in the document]
cammellos: mainly consistency, messageID has been hex-encoded so far, and it's hard to keep track of things if we mix the two, given that also status-react generates its ids for some messages (local never sent for example), having it like this is one less thing to think about, so I'd rather keep it, but if you feel strongly that should be `[]byte` I don't mind changing it

encryption/encryptor.go
```diff
@@ -123,7 +122,8 @@ func (s *encryptor) ConfirmMessageProcessed(messageID []byte) error {
 	confirmationData, ok := s.messageIDs[id]
 	if !ok {
 		s.logger.Debug("could not confirm message", zap.String("messageID", id))
-		return fmt.Errorf("message with ID %#x not found", messageID)
+		// We are ok with this, means no key material is stored (public message, or already confirmed)
```
PombeirP: Should the log message reflect it then (e.g. `could not confirm message or message already confirmed`, or something like that)?

encryption/protocol.go
```diff
@@ -405,6 +405,8 @@ func (p *Protocol) GetPublicBundle(theirIdentityKey *ecdsa.PublicKey) (*Bundle,
 
 // ConfirmMessageProcessed confirms and deletes message keys for the given messages
 func (p *Protocol) ConfirmMessageProcessed(messageID []byte) error {
+	logger := p.logger.With(zap.String("site", "ConfirmMessageProcessed"))
+	logger.Debug("confirm message", zap.Binary("message-id", messageID))
```
PombeirP: Shouldn't this read like this?

```suggestion
	logger.Debug("message confirmed", zap.Binary("message-id", messageID))
```

messenger_test.go
```diff
@@ -347,9 +347,184 @@ func (s *MessengerSuite) TestChatPersistencePrivateGroupChat() {
...
+	s.Require().Equal(1, len(savedContacts))
+	s.Require().Equal("blocked", savedContacts[0].Name)
+
+	// The chat is deleted
+	actualChats, err := s.m.Chats(0, -1)
+	s.Require().NoError(err)
+	s.Require().Equal(2, len(actualChats))
+
+	// The messages have been deleted
+	chat1Messages, _, err := s.m.MessageByChatID(chat2.ID, "", 20)
```
PombeirP: The returned variable is called `chat1Messages` while we're asking for messages from `chat2`

messenger_test.go
```diff
@@ -347,9 +347,184 @@ func (s *MessengerSuite) TestChatPersistencePrivateGroupChat() {
...
+	s.Require().NoError(err)
+	s.Require().Equal(3, len(chat1Messages))
+
+	chat2Messages, _, err := s.m.MessageByChatID(chat3.ID, "", 20)
```
PombeirP: Same here

persistence.go
```diff
@@ -93,7 +86,7 @@ func (db sqlitePersistence) SaveChat(chat Chat) error {
 	defer stmt.Close()
 
 	_, err = stmt.Exec(
-		chatID,
+		chat.ID,
```
PombeirP: Don't we want to do some kind of argument validation on method entry?

persistence_legacy.go
```diff
@@ -271,3 +320,61 @@ func (db sqlitePersistence) UpdateMessageOutgoingStatus(id string, newOutgoingSt
...
+func (db sqlitePersistence) BlockContact(contact Contact) ([]*Chat, error) {
+	var (
+		tx  *sql.Tx
+		err error
+	)
+	tx, err = db.db.BeginTx(context.Background(), &sql.TxOptions{})
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
```
PombeirP: Maybe we could write a simple helper function that commits or rollbacks a tx based on received err, tx arguments?

cammellos: I might do a separate PR refactoring once the code settles a bit, but good idea

persistence.go
cammellos: probably, maybe though I will leave for a separate PR, we might want to think about what kind of validation and where exactly we want to have it

persistence_legacy.go
```diff
@@ -183,61 +247,41 @@ func (db sqlitePersistence) MessageByChatID(chatID string, currCursor string, li
...
 
-func (db sqlitePersistence) UnseenMessageIDs() ([][]byte, error) {
-	rows, err := db.db.Query(`SELECT id FROM user_messages_legacy WHERE seen = 0`)
+	stmt, err = tx.Prepare(query)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	defer rows.Close()
+	defer func() {
```
adambabik: This should be right after checking error for `tx, err = ...`. I think that's the place where a tx is created. If `tx.Prepare()` returns an error, the tx will be never completed.

v1/message.go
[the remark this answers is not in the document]
adambabik: Sounds good. Just wanted to be sure the it's actually simpler and it sounds that it is.

PR #52, opened 2019-08-20
Use wrapped payload for id calculation

cammellos:
Now that status-react is not calculating ids anymore (only status-protocol-go does), we can use the wrapped payload for id calculation also when receiving.

### DIFF
--- a/v1/status_message.go
+++ b/v1/status_message.go
@@ -125,8 +125,9 @@ func (m *StatusMessage) HandleApplicationMetadata() error {
 		return err
 	}
 	m.ApplicationMetadataLayerSigPubKey = recoveredKey
+	// Calculate ID using the wrapped record
+	m.ID = MessageID(recoveredKey, m.DecryptedPayload)
 	m.DecryptedPayload = message.Payload
-	m.ID = MessageID(m.SigPubKey(), m.DecryptedPayload)
 	return nil
 
 }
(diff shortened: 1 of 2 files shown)


v1/status_message.go
```diff
@@ -125,8 +125,9 @@ func (m *StatusMessage) HandleApplicationMetadata() error {
 		return err
 	}
 	m.ApplicationMetadataLayerSigPubKey = recoveredKey
-	m.DecryptedPayload = message.Payload
+	// Calculate ID using the wrapped record
 	m.ID = MessageID(m.SigPubKey(), m.DecryptedPayload)
+	m.DecryptedPayload = message.Payload
```
adambabik: It looks like we can calculate the ID in the first line of this method or even in the method handling datasync layer.

cammellos:
Is fairly similar, the advantage would be that we can change any field in the wrapper without breaking id calculation, so we have slightly more freedom to change fields around

v1/status_message.go
cammellos: actually, we can't, because it uses the signature of the message, I have made that more explicit now by using `recoveredKey` directly

adambabik: Ah it was hidden by `m.SigPubKey()` which took a value from `ApplicationMetadataLayerSigPubKey` field in this line. Now it's clear :)

PR #54, opened 2019-08-21
Add persistent datasync

cammellos:
Add datasync persistence


adambabik:
For the custom migrations table we need to merge https://github.com/status-im/migrate/pull/2.

cammellos:
@adambabik could you review again, I have made some changes, basically, I have exposed a method
`Migrate` in mvds which takes care of migrating the db passed with the corresponding migrations.
In turn this is called by our `Migrate` method, which migrates the passed db with all the correct migrations + mvds, so that the two migrations are not stored in the same table.

sqlite/db.go
```diff
@@ -117,3 +117,21 @@ func ApplyMigrations(db *sql.DB, assetNames []string, assetGetter func(name stri
 
 	return nil
 }
+
+func Migrate(database *sql.DB) error {
```
adambabik: Hm, when I look at this now, I think we can keep `sqlite` package generic and move this method along with `prepareMigrations` to the root project directory. What do you think? 

Additionally, `ApplyMigrations` would need to take `sqlcipher.Config` as an argument.

We can also think about it in a separate PR.

cammellos: Not sure, I personally prefer to have all this abstracted from the rest of the app, i.e, I only pass a `db` and don't have to worry about passing specific migrations/config to it, as currently I don't see a use case for passing different migrations (other than testing that particular migration in isolation I guess, but I'd rather test them all together), the only use cases I see are: 
1) Open a new database, apply migrations (Open method) 
2) Migrate an existing database `Migrate`

But I might be missing something. In any case if you are ok delaying the decision I'd rather do it in a separate PR, I would like to get this through testing to make sure all 3 migrations (status-go/protocol/mdvs work ok together with a single database)

adambabik: That's not exactly what I mean. I am down for `func Migrate(database *sql.DB) error`, I would just put it in the root directory. From the caller perspective nothing changes and `sqlite` package does not know about its sibling packages.

Feel free to merge as it is though. We can refactor later if it still makes sense.

cammellos: oh I see, I misunderstood, yes I have no problem with that, I might refactor in the next pr, just to get this through, but sounds good

PR #55, opened 2019-08-21
add Messenger.Init()

adambabik:
This change adds `Messenger.Init()` method which based on the stored chats and contacts initializes Whisper filters.

It also contains some changes cleaning up term `filter` and `chat`.

Todo:
- [ ] add unit tests for `Messenger.Init()`

### DIFF
--- a/contact.go
+++ b/contact.go
@@ -1,5 +1,18 @@
 package statusproto
 
+import (
+	"crypto/ecdsa"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+const (
+	contactBlocked         = "contact/blocked"
+	contactAdded           = "contact/added"
+	contactRequestReceived = "contact/request-received"
+)
+
 // ContactDeviceInfo is a struct containing information about a particular device owned by a contact
 type ContactDeviceInfo struct {
 	// The installation id of the device
@@ -13,7 +26,7 @@ type ContactDeviceInfo struct {
 // Contact has information about a "Contact". A contact is not necessarily one
 // that we added or added us, that's based on SystemTags.
 type Contact struct {
-	// ID of the contact
+	// ID of the contact. It's a hex-encoded public key (prefixed with 0x).
 	ID string `json:"id"`
 	// Ethereum address of the contact
 	Address string `json:"address"`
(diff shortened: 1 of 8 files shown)


messenger.go
```diff
@@ -317,6 +318,55 @@ func NewMessenger(
...
+	}
+	for _, chat := range chats {
+		if !chat.Active {
+			continue
+		}
+		if chat.ChatType == ChatTypePublic {
+			publicChatIDs = append(publicChatIDs, chat.ID)
+		} else if chat.ChatType == ChatTypeOneToOne {
+			publicKeys = append(publicKeys, chat.PublicKey)
+		}
```
cammellos: Forgot to mention sorry, we also need to create a filter for each group chat member, `ChatMember.ID` is the hex encoded public key

contact.go
```diff
@@ -31,3 +44,33 @@ type Contact struct {
...
+
+func (c Contact) HasBeenAdded() bool {
+	return existsInStringSlice(c.SystemTags, contactRequestReceived)
+}
+
+func (c Contact) IsBlocked() bool {
+	return existsInStringSlice(c.SystemTags, contactBlocked)
+}
+
+// exists checks if a string is in a set.
```
PombeirP:
```suggestion
// existsInStringSlice checks if a string is in a set.
```

messenger.go
```diff
@@ -317,6 +318,55 @@ func NewMessenger(
...
+	// TODO: Get only active chats by the query.
+	chats, err := m.Chats()
+	if err != nil {
+		return err
+	}
+	for _, chat := range chats {
+		if !chat.Active {
+			continue
+		}
+		if chat.ChatType == ChatTypePublic {
```
PombeirP: Why not use a `switch` statement instead?

messenger.go
adambabik: Added.

adambabik:
Sorry, I forgot about unit tests. Will add asap.

messenger.go
cammellos: looks good, thanks, just a note that you might have duplicates, but if we are fine with that and they are handled properly then no probs

adambabik: It is. If a filter of the same type for a given public key exists, it will be returned early without creating a duplicate.

PR #57, opened 2019-08-22
Remove sqlite persistence

cammellos:
Currently there seem to be a bug with publishing the contact-code, and
each attempt result in the error `database is closed` (non-thread safe?). As discussed
previously we probably don't need to have fs persistence, so just
changing this to have it in memory.


cammellos:
haven't quite looked into it as we already discussed to just use in-memory, but I might have an extra look, thanks!

PR #66, opened 2019-08-29
Improve interface around Chat

adambabik:
* Provide functions to create `Chat` instances
* `Send*` methods expect `chatID` instead of `Chat` instance now
* separates some responsibilities


messenger.go
```diff
@@ -697,3 +724,129 @@ func (m *Messenger) MarkMessagesSeen(ids ...string) error {
...
+		chatID := message.Content.ChatID
+		chat := findChatByID(chatID, chats)
+		if chat == nil {
+			return nil, errors.New("received a public message from non-existing chat")
+		}
+		return chat, nil
+	case message.MessageT == protocol.MessageTypePrivate && isPubKeyEqual(message.SigPubKey, p.myPublicKey):
+		// It's a private message coming from us so we rely on Message.Content.ChatID.
+		chatID := message.Content.ChatID
+		chat := findChatByID(chatID, chats)
```
cammellos: Not sure this is correct, the behavior should be that if the chat does not exist it should be created in this case. Say I send a message to you from device A, I will send the same message to device B. Device B should create a chat similarly to the next case.

PR #74, opened 2019-09-18
Port identicon & gfycat to status

cammellos:
Adds two fields, generatedName ( I will rename to gfycat maybe? ) and Identicon

code for identicon taken from https://github.com/jgimeno/go-identicon , slightly modified


messenger.go
```diff
@@ -697,3 +740,13 @@ func (m *Messenger) MarkMessagesSeen(ids ...string) error {
 func (m *Messenger) UpdateMessageOutgoingStatus(id, newOutgoingStatus string) error {
 	return m.persistence.UpdateMessageOutgoingStatus(id, newOutgoingStatus)
 }
+
+func Identicon(id string) string {
+	return identicon.GenerateBase64(id)
+}
+
+func GenerateName(id string) string {
```
adambabik: Not used?

cammellos: This one we use in `status-react` (same for identicon), because generation is only in here, we need to export to status-react, as it is used synchronously to generate names (as it was before), otherwise it would require too many changes in one go.

gfycat/generate.go
```diff
@@ -0,0 +1,41 @@
...
+	animal := animals[animalIndex]
+
+	return fmt.Sprintf("%s %s %s", adjective1, adjective2, animal)
+}
+
+func GenerateFromPublicKey(publicKey *ecdsa.PublicKey) string {
+	return Generate(uint64(publicKey.X.Int64()))
+}
+
+func GenerateFromPublicKeyString(publicKeyString string) (string, error) {
```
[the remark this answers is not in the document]
cammellos: I can have the caller pass an `ecdsa.PublicKey` instead, but then I will have at least two places where I have to make the conversion (both in messenger, `SaveContact` and `GenerateName`), I don't mind doing it, but not sure it's cleaner

cammellos:
> let's make sure we handle errors
Done

> let's add comments at least for exported methods
Done

> why is it called gfycat? :D

I think gfycat popularized it https://medium.com/@Gfycat/naming-conventions-97960fc40179 , not sure what a better name would be

> let's make sure we export only what's needed so that we don't need to maintain huge public interface

Done

jakubgs:
>not sure what a better name would be

Agree with Adam. Names that require obscure context are bad. Something generic like `identity`, and put the icon module in there too under `icon`, so you get `identity/icon` and `identity/username` or something like that.

cammellos:
GeneratedIcon?

gfycat/generate.go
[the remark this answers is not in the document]
adambabik: Comment makes it clear 👍 let's keep it then.

cammellos:
@adambabik @jakubgs I have moved:
`identity/username` -> 3 random words
`identity/identicon` -> Picture

And the corresponding fields:
`username` and `identicon` in contact. 

There some potential confusion with `name` which is set by the user, so it kind of a username as well, but I think we are getting rid of it eventually, so should not be a problem.
Let me know what you think

messenger.go
```diff
@@ -659,13 +678,52 @@ func (m *Messenger) RetrieveRawAll() (map[transport.Filter][]*protocol.StatusMes
...
+
+			identicon, err := identicon.GenerateBase64(id)
+			if err != nil {
+				continue
+			}
+
+			contact := Contact{
+				ID:        id,
+				Address:   address[2:],
+				Username:  message.GenerateUsername(),
```
```suggestion
				Username:  username.GenerateFromPublic(message.SigPubKey()),
```

v1/status_message.go
```diff
@@ -47,6 +50,17 @@ func (s *StatusMessage) SigPubKey() *ecdsa.PublicKey {
 	return s.TransportLayerSigPubKey
 }
 
+// GenerateUsername returns the 3 words name of the most important signature
+func (s *StatusMessage) GenerateUsername() string {
```
adambabik: Seems unused. I am not convinced `StatusMessage` needs to know about `identicon` and `username` packages. We only do that because we have a public key for each message but username and identicon are properties of contact.

cammellos: We use them in status-go, each message comes with a `Author` with `Name` and `Identicon`, I don't mind though removing the method and have it calculated in `status-go` using `SigPubKey`, similar calculations will have to be done for each client though.

jakubgs:
>There some potential confusion with name which is set by the user, so it kind of a username as well

That's a good point. I guess it is a bit close. Maybe `pseudonim` or `alias` would be closer to the menaing?

cammellos:
@jakubgs alias sounds reasonable, I might change it to that if there's not any objections, @adambabik ?

v1/status_message.go
adambabik: We also have these exposed as functions in `messenger.go` so it can be delegated to them. Would that be ok? 

I am trying to limit dependencies between packages as much as possible, especially here to keep `v1` independent. Also, `v1` as a package probably wasn't the best idea...

cammellos: yes, that would work fine, we can just use those to attach he data, will do that

jakubgs:
Is this ready to merge or is there anything else to be done here @adambabik ?

cammellos:
@jakubgs it needs to go through testing in status-react, I'd rather get a green light there and then merge, otherwise if there's anything to fix I will have to open multiple prs etc

jakubgs:
Yeah, makes sense, just wanted to know what's left to be done, thanks!

Issue #73, opened 2019-09-17
Port group messages logic


adambabik:
As a developer, I would like to be able to send group messages using the status-protocol-go so that I can communicate with other clients.

The current implementation is in [status-react](https://github.com/status-im/status-react/blob/e87d75039ead5a1baf77daeb657be489d442644a/src/status_im/group_chats/core.cljs).

#### Acceptance criteria

- in `v1/` add `CreatePrivateGroupTextMessage` function
- extend `messageProcessor` with `SendGroup` method
- make sure `messageProcessor.Process` properly handles group messages
- port membership updates validation logic
- port sending membership updates logic
- handle membership updates logic
- in Whisper transport add methods to set up proper topic filters
- in `Messenger.Init` setup proper filters based on active group chats

cc @cammellos

adambabik:
Payload example: https://gist.github.com/cammellos/8621121434b2a470ff0fe1f3443929b2

PR #78, opened 2019-09-24
Implement signing MembershipUpdate message

adambabik:
Relates to #73


v1/membership_update_message_test.go
```diff
@@ -48,3 +49,22 @@ func TestEncodeMembershipUpdateMessage(t *testing.T) {
...
+		{"type", "chat-created"},
+	}, result)
+
+	key, err := crypto.GenerateKey()
+	require.NoError(t, err)
+	update := testMembershipUpdateMessageStruct.Updates[0]
+	err = SignMembershipUpdate(&update, key)
+	require.NoError(t, err)
+
+	// TODO: we should generate a signature in ClojureScript and
```
adambabik: @cammellos would it be easy to generate? I am thinking of sharing a private key, signature and events between this lib and status-react. This way we can ensure in unit tests that both implementations are correct and prevent future incompatibilities.

cammellos: Not sure I understand exactly, I can generate probably a signature for a given private key, but:

Currently the signing happens in `status-go`: https://github.com/status-im/status-go/blob/develop/mobile/status.go#L88

we pass a JSON formatted string from `status-react` to `status-go` that will be signed:
`"[[[....]]]'`

So I would have only one part of the codebase to generate that (`status-protocol-go`),
to increase the surface of shared code (so that we slowly move to use status-protocol-go for private group-chats), I would pass instead of a string a golang struct, `MembershipUpdate` (which should be compatible), and just have one place we do the signing.

This way we have a single place, so `status-react` and `status-protocol-go` won't have to be kept synchronized in the way the sign messages (only `status-protocol-go` will do.). Also no need to make sure the signature is identical with the one generated by status-react/status-go (although they should be, and if not we should understand why they are not), as we don't need to keep backward compatibility.

What do you think? did I understand what you mean?

adambabik: That totally makes sense. I was worried about the serialization to a string `[[[...]]]` and keeping compatibility with the current codebase. But, as you pointed out that backward compatibility is not an issue, we should concentrate all codebase in this lib and it won't be a problem.

No action needed :) Thanks!

v1/membership_update_message.go
```diff
@@ -35,3 +99,64 @@ func EncodeMembershipUpdateMessage(value MembershipUpdateMessage) ([]byte, error
 	}
 	return buf.Bytes(), nil
 }
+
+// SignMembershipUpdate signs a slice of MembershipUpdateEvents
+// and updates MembershipUpdate's signature.
+// It follows the algorithm describe in the spec: https://github.com/status-im/specs/blob/master/status-group-chats-spec.md#signature.
+func SignMembershipUpdate(update *MembershipUpdate, identity *ecdsa.PrivateKey) error {
```
adambabik: This will be a method on `MembershipUpdate` in the next PR.

v1/membership_update_message.go
```diff
@@ -26,6 +34,62 @@ type MembershipUpdateEvent struct {
 	Name       string   `json:"name,omitempty"`    // name of the group chat
 }
 
+func NewChatCreatedEvent(name string, clock int64) MembershipUpdateEvent {
+	return MembershipUpdateEvent{
+		Type:       "chat-created",
```
adambabik: I moved types of events to consts in the next PR.

PR #80, opened 2019-09-26
Revert "Make database optional (#72)"

cammellos:
This reverts commit 3fa3648db96d9660c41e4aeed7f26f70a5e1f773.

Seems like the app is now using an in memory database, reverting this for now as likely the cause (I will verify it before merging), if the fix is easy enough feel free to appy, but there's a pending pr that I need to merge and has already tested successfully.

### DIFF
--- a/messenger.go
+++ b/messenger.go
@@ -235,25 +235,22 @@ func NewMessenger(
 		}
 	}
 
-	var err error
-
 	// Configure the database.
 	database := c.db
-	if database == nil && c.dbConfig != (dbConfig{}) {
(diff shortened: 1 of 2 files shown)


cammellos:
@adambabik it's using in memory yes, but I haven't changed any config in status-go/status-react, so just by using this version, it uses an in memory database, with the reverted commit (and nothing else changed), it uses the fs

cammellos:
@adambabik one more review please https://github.com/status-im/status-go/pull/1619 :)

PR #86, opened 2019-10-03
Implement group messages logic

adambabik:
These changes implement logic of validating and handling new updates to the group.

Each update is verified, the events are validated and recomputed, and the chat is updated. 

This PR contains one backward incompatible change. The signature of `Messenger.Send` changes to:
```go
Send(ctx context.Context, chatID string, data []byte) ([][]byte, error)
```
because now `Send` can send multiple messages if a group chat has more than 1 member.

## Limitations

- leaving group will also remove 1-1 chat, don't use :)
- no handling of members removal
- no managing group admins


transport/whisper/whisper_service.go
```diff
@@ -185,6 +188,30 @@ func (a *WhisperServiceTransport) LeavePrivate(publicKey *ecdsa.PublicKey) error
...
+	for _, pk := range publicKeys {
+		_, err = a.filters.LoadContactCode(pk)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (a *WhisperServiceTransport) LeaveGroup(publicKeys []*ecdsa.PublicKey) error {
```
cammellos: Not sure we can do that, you can remove filters only if:
1) User is not in our contacts
2) We don't have any 1-to-1 chat with them
3) It's not in any group

Basically when deciding to remove filters for a user, you need to check that there are no other reason to have those filters up (also negotiated filters should never be removed, unless banning, so only contact-code should be removed)

Or am I missing something?

adambabik: No, you're totally right. I commented about it here https://github.com/status-im/status-protocol-go/pull/86/files#diff-8d8b2746f48de4eb1c3158c07d06d19bR84-R86. I will add to the description the current limitations. There are more, for example, no way to remove members and no managing admins. I will add these in another PR.

PR #88, opened 2019-10-04
Prepare API to support other implementations of ETH nodes

### LINKED ISSUE #84: Prepare API for alternative ETH node implementations
Currently, there are areas in the code that leak Geth types, making it impossible to have status-protocol-go work with e.g. Nimbus.

### Issues

- [x] `NewMessenger` is receiving a Geth type `*whisper.Whisper`;
- [x] `whisper.BytesToTopic` needs to be encapsulated;
- [x] `WhisperServiceTransport.RetrieveAllMessages` is returning a Geth type (`*whisper.ReceivedMessage`), tying the consuming class `Messenger` to Geth;
- [x] `messageProcessor.Process` receives `*whisper.ReceivedMessage`.

### Implementation

To fix that we need to create interfaces/wrappers for:
  - [x] `whisper.Whisper`
  - [x] `whisper.PublicWhisperAPI`
  - [x] `whisper.Filter`
    - [x] Factory method on Whisper interface that creates appropriate `Filter` wrapper object instances
  - [x] `whisper.NewMessage`
  - [x] `whisper.ReceivedMessage`
  - [x] `whisper.MessageStore`
  - [x] `whisper.BytesToTopic`
  - [x] Fix tests
  - Optional:
    - [x] `envelopesMonitorConfig`

### Next steps

- Implement Nimbus support
- Have status-console-client work on top of Nimbus

cc @adambabik @cammellos

PombeirP:
Closes #84 

This PR tries to limit usage of Geth types to the strict minimum surface area, and instead expose its own types, or wrapper interfaces to the outside world so that other implementations like Nimbus can more easily be added.

For the time being, the envelope monitor has been excluded from the scope (commented out), since it is not strictly required for the MVP. Otherwise, it is a non-breaking change.

This PR is part of a group of 3 PRs (https://github.com/status-im/status-protocol-go/pull/88, https://github.com/status-im/status-go/pull/1627, https://github.com/status-im/status-console-client/pull/128) and we should only merge when we're happy with the 3 of them.


transport/whisper/envelopes.go
```diff
@@ -27,65 +25,70 @@ type EnvelopesMonitorConfig struct {
...
 	Logger                         *zap.Logger
 }
 
 // EnvelopeEventsHandler used for two different event types.
 type EnvelopeEventsHandler interface {
 	EnvelopeSent([][]byte)
 	EnvelopeExpired([][]byte, error)
-	MailServerRequestCompleted(common.Hash, common.Hash, []byte, error)
-	MailServerRequestExpired(common.Hash)
+	MailServerRequestCompleted(whispertypes.Hash, whispertypes.Hash, []byte, error)
```
adambabik: `Hash` is not a Whisper-only type. It's used thoroughly. We can replace it with `[]byte` probably if we need to?

adambabik:
I don't exactly see how it's gonna work with two implementations. 

So, the idea is that nim-based implementation, which would interact with Go using Cgo, is gonna implement proposed interfaces and wrappers by kind of translating Cgo code to Go and vice versa? Maybe we can have some diagram or some snippet that would explain how this effort is solving the problem of multiple implementation? My main point is to introduce some more clarity and further verify assumptions (maybe they were already verified but I can't find any trace of that in this PR or a linked issue).

message_processor.go
```diff
@@ -201,14 +201,14 @@ func (p *messageProcessor) SendPublic(ctx context.Context, chatID string, data [
...
-	var newMessage whisper.NewMessage
+	var newMessage whispertypes.NewMessage
 
 	wrappedMessage, err := p.tryWrapMessageV1(data)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to wrap message")
 	}
 
-	newMessage = whisper.NewMessage{
+	newMessage = whispertypes.NewMessage{
```
kdeme: Do we need to have this message created at this level? TTL, PoW and PoWTime remain always the same, no? Just passing a payload would make this more transport agnostic.

kdeme:
Update: OK, after going through it a bit further and seeing the wrapper go files I see that you are indeed facilitating to swap at Whisper API level, sorry for ignorance :)

So it would e.g. be swapped at `return a.shhAPI.Post(ctx, *newMessage)`, the `Post` call, right?

transport/whisper/types/filter.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Filter represents a Whisper message filter
+type Filter interface {
+	Src() *ecdsa.PublicKey      // Sender of the message
+	KeyAsym() *ecdsa.PrivateKey // Private Key of recipient
+	KeySym() []byte             // Key associated with the Topic
+	Topics() [][]byte           // Topics to filter messages with
+	PoW() float64               // Proof of work as described in the Whisper spec
+	AllowP2P() bool             // Indicates whether this filter is interested in direct peer-to-peer messages
+	SymKeyHash() Hash           // The Keccak256Hash of the symmetric key, needed for optimization
```
kdeme: Why is `SymKeyHash` here? Should this not be a whisper implementation specific internal?

transport/whisper/types/received_message.go
```diff
@@ -0,0 +1,30 @@
+package whispertypes
+
+import (
+	"crypto/ecdsa"
+)
+
+// ReceivedMessage represents a data packet to be received through the
+// Whisper protocol and successfully decrypted.
+type ReceivedMessage interface {
+	Raw() []byte
```
kdeme: Is `Raw()` here the full raw message? Why is that needed?
Is `Signature()` needed? After all there is already `Src` which gets recovered from the signature.
Which salt is `Salt()` returning?

Furthermore, in the Nimbus Whisper implementation we do not expose `P2P` and `SymKeyHash`. Are both needed?
Important side note that I haven't mentioned before: We do not have a mailserver implementation at the moment.

transport/whisper/types/whisper.go
```diff
@@ -0,0 +1,61 @@
...
+	// SelectKeyPair adds cryptographic identity, and makes sure
+	// that it is the only private key known to the node.
+	SelectKeyPair(key *ecdsa.PrivateKey) error
+	AddSymKeyDirect(key []byte) (string, error)
+	AddSymKeyFromPassword(password string) (string, error)
+	DeleteSymKey(id string) bool
+	GetSymKey(id string) ([]byte, error)
+
+	Subscribe(f Filter) (string, error)
+	GetFilter(id string) Filter
```
kdeme: The Nimbus implementation already has a `getFilterMessages` implemented. You retrieve messages through the filterId and not by passing the actual filter. The API only offers this call to retrieve messages, or via a callback. This way there is no need for a `GetFilter()` call.

messenger_test.go
```diff
@@ -54,8 +56,9 @@ func (s *MessengerSuite) SetupTest() {
 
 	config := whisper.DefaultConfig
 	config.MinimumAcceptedPOW = 0
-	s.shh = whisper.New(&config)
-	s.Require().NoError(s.shh.Start(nil))
+	shh := whisper.New(&config)
+	s.shh = gethbridge.NewGethWhisperWrapper(shh)
+	s.Require().NoError(shh.Start(nil))
```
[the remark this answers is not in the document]
PombeirP: I don't see a special reason to do it. Starting a node is implementation-specific, or not?

transport/whisper/envelopes.go
PombeirP: I just had to create a folder in `/types` for `HexBytes` (to replace `hexutils.Bytes`), I guess I could now move this there?

PombeirP:
> So it would e.g. the `Post` call would be swapped at `return a.shhAPI.Post(ctx, *newMessage)`, right?
> And then specific `func (w *gethPublicWhisperAPIWrapper) Post` should be reimplemented in the Nimbus version?

Exactly

message_processor.go
PombeirP: I agree, but my goal with this PR is to make a first pass at getting rid of Geth-specific types and still having something that works. Baby steps. Also, you shouldn't forget that just because that approach would work well in this particular function, there are other places that still need a Geth-agnostic `NewMessage` struct. E.g. in the [`EnvelopesMonitor` class](https://github.com/status-im/status-protocol-go/blob/feature%2Fstatus-nim/transport/whisper/envelopes.go#L64), which will work over either version of the node.

transport/whisper/types/filter.go
PombeirP: To be honest, I did a mostly automated search and replace over the structs, so everything is present here. Now that everything is building and working again, I can go over these methods and remove the ones which are never used.

transport/whisper/types/received_message.go
PombeirP: See explanation above. Will be removing unused functions next.

transport/whisper/types/whisper.go
PombeirP: That's a good point. `status-protocol-go` seems to be using `GetFilter` (e.g. [here](https://github.com/status-im/status-protocol-go/blob/feature%2Fstatus-nim/transport/whisper/whisper_service.go#L196)), but I can probably refactor that to go through `PublicWhisperAPI.GetFilterMessages`.

PombeirP: I replaced all but one `GetFilter` method calls, because the status-protocol-go method is using it to return a `map[Filter][]whispertypes.ReceivedMessage`. That method is marked as deprecated (probably being used still by status-react), so we will be able to drop it.

PombeirP: @kdeme One hitch here is that `Filter.Retrieve` returns `whisper.RetrievedMessages` while `GetFilterMessages` returns a `whisper.Message` (which is the RPC representation).

message_processor.go
kdeme: Yeah, understood. Not really in scope of this PR. And I realize that while here it is easy to move, at other locations it will be more difficult to get transport agnostic.

transport/whisper/types/received_message.go
kdeme: I see a lot got removed, but OK if that is all what is required.

transport/whisper/types/whisper.go
kdeme: So considering you now replace all `Retrieve` with `GetFilterMessages`. Can't `RetrievedMessage` go away completely? Or do we still need the `ToWhisperMessage` call?

kdeme:
Looks good to me. I think any message or other remapping that needs to be done for Nimbus can now be done in the specific wrapper (or we can always do some minor changes to the Nimbus API itself).

While I think it will still be important that in the future the separation between the transport (Whisper) and the Status protocol (and then specifically here the concept of Public, private, etc. chats) is improved, I agree that it is not in scope of this PR. So ignore those comments for now.

transport/whisper/types/whisper.go
PombeirP: Yeahh, I thought the same (and actually removed `ToWhisperMessage`), but we still need the type [here](https://github.com/status-im/status-protocol-go/blob/60301521af104e95579e38a61e361eb84350f454/transport/whisper/gethbridge/message_store_wrapper.go#L29)

PombeirP:
OK, ready for final review.

messenger_test.go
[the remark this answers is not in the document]
adambabik: I don't think so. Assume one wants to create the bridge and pass a struct around but not necessarily start Whisper yet. With this setup, it means both the wrapper and `shh` need to be kept in that struct.

transport/whisper/envelopes.go
```diff
@@ -128,12 +132,13 @@ func (m *EnvelopesMonitor) GetState(hash common.Hash) EnvelopeState {
...
 // handleEnvelopeEvents processes whisper envelope events
 func (m *EnvelopesMonitor) handleEnvelopeEvents() {
-	events := make(chan whisper.EnvelopeEvent, 100) // must be buffered to prevent blocking whisper
+	events := make(chan whispertypes.EnvelopeEvent, 100) // must be buffered to prevent blocking whisper
 	sub := m.w.SubscribeEnvelopeEvents(events)
 	defer sub.Unsubscribe()
 	for {
 		select {
 		case <-m.quit:
+			close(events)
```
adambabik: Can we close before unsubscribing? We also don't handle `sub.Error` I think.

PombeirP: Do you have a link? I don't know what `sub.Error` you're referring to.

adambabik: `sub` is an interface https://sourcegraph.com/github.com/status-im/whisper@3e303b1d5fc15a373594b56f7ca224ebc5b3c4ca/-/blob/vendor/github.com/ethereum/go-ethereum/event/subscription.go#L41-45.

messenger_test.go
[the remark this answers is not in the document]
PombeirP: Maybe I'm not picturing what you mean correctly. Would that mean that we'd also wrap the `p2p.Server` struct that is used as argument, to be used later on when the bridge would start the node? The goal here is to avoid leaking those types from this API.

transport/whisper/envelopes.go
PombeirP: what do we want to do in that scenario? Just return like this?

``` go
case <-sub.Err():
	return
```

messenger_test.go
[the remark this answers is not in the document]
adambabik: My idea is as simple as having a possibility to do:
```go
s.shh.Start()
```

But I forgot about `p2p.Server` as an argument... That would require to pass `p2p.Server` to `whisper.New` constructor so that we don't need to wrap `p2p.Server`. However, that's even more complex change so let's keep it as it is.

transport/whisper/envelopes.go
adambabik: I guess so. We don't handle errors in any particular way here.

adambabik: But it's also super minor. We can keep it as it is.

PombeirP:
Are we OK with merging this PR then?

PombeirP:
@cammellos @adambabik @kdeme By the way, I think the future goal of `status-protocol-go` should be to not include neither geth nor Nimbus, but rather those should be adapter projects (e.g. `status-transport-geth` and `status-transport-nimbus`) that get plugged in to `status-protocol-go`. These wrappers should then be moved to the respective adapter projects.

PR #91, opened 2019-10-10
Add ens name verification

cammellos:
Add ENS verification to status-protocol-go.

I have added a test, but is only run if `RPC_ENDPOINT` is set in the enviroment variable, that's because is fairly flaky due to hitting the quota limit + it's an online test, but it's still valuable imo as it check e2e. 

Also fixes a few issues with the move to different types;
1) JSON was not marshalled correctly as some of the fields used hexutil.bytes
2) Re-introduced method used by status-react, but using the new types now.


go.mod
```diff
@@ -19,11 +22,12 @@ require (
...
+	github.com/vacp2p/mvds v0.0.0-20191009025501-899aeeacb474
+	github.com/wealdtech/go-ens/v3 v3.0.7
 	go.uber.org/zap v1.10.0
 	golang.org/x/crypto v0.0.0-20190701094942-4def268fd1a4
 	golang.org/x/net v0.0.0-20190628185345-da137c7871d7 // indirect
 	golang.org/x/sys v0.0.0-20190801041406-cbf593c0f2f3 // indirect
 )
 
-replace github.com/ethereum/go-ethereum v1.9.5 => github.com/status-im/go-ethereum v1.9.5-status.4
+replace github.com/ethereum/go-ethereum v1.9.2 => github.com/status-im/go-ethereum v1.9.5-status.4
```
adambabik: Something weird happened here. Why did it change to `1.9.2`?

messenger.go
```diff
@@ -690,6 +699,26 @@ func (m *Messenger) RetrieveRawAll() (map[transport.Filter][]*protocol.StatusMes
 	return result, nil
 }
 
+func buildContact(publicKey *ecdsa.PublicKey) (*Contact, error) {
+	address := strings.ToLower(crypto.PubkeyToAddress(*publicKey).Hex())
+
+	id := fmt.Sprintf("0x%s", hex.EncodeToString(crypto.FromECDSAPub(publicKey)))
```
```suggestion
	id := "0x" + hex.EncodeToString(crypto.FromECDSAPub(publicKey))
```

Or this new `hexBytes` struct.

messenger.go
```diff
@@ -690,6 +699,26 @@ func (m *Messenger) RetrieveRawAll() (map[transport.Filter][]*protocol.StatusMes
 	return result, nil
 }
 
+func buildContact(publicKey *ecdsa.PublicKey) (*Contact, error) {
```
adambabik: Can we put it into `contact.go`?

messenger.go
```diff
@@ -916,6 +935,111 @@ func (p *postProcessor) matchMessage(message *protocol.Message, chats []*Chat) (
 	}
 }
 
+type EnsDetails struct {
```
adambabik: 1. IMO it should be a separate file at least. Like `ens.go`.
2. Small detail but `ENSDetails` and `ENSRespoonse` is a recommended way:
```
Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url". Here's an example: ServeHTTP not ServeHttp.

This rule also applies to "ID" when it is short for "identifier," so write "appID" instead of "appId".
```

messenger.go
```diff
@@ -916,6 +935,111 @@ func (p *postProcessor) matchMessage(message *protocol.Message, chats []*Chat) (
...
+	// Assemble the bytes returned for the pubkey
+	pubKeyBytes := elliptic.Marshal(crypto.S256(), new(big.Int).SetBytes(x[:]), new(big.Int).SetBytes(y[:]))
+
+	response.PublicKey = publicKey
+	response.Verified = bytes.Equal(pubKeyBytes, expectedPubKeyBytes)
+	output <- response
+}
+
+// VerifyENSName verifies that a registered ENS name matches the expected public key
+func (m *Messenger) VerifyENSNames(ensDetails []EnsDetails, rpcEndpoint, contractAddress string) (map[string]EnsResponse, error) {
```
```suggestion
func (m *Messenger) VerifyENSNames(rpcEndpoint, contractAddress string, ensDetails []EnsDetails) (map[string]EnsResponse, error) {
    // Can address and RPC endpoint have some defaults? Does it change during runtime? 
    verifier := ENSVerifier(WithContractAddress(contractAddress), WithRPCEndpoint(rcpEndpoint))
    response, err := verifier.CheckBatch(ensDetails)
    ...
}
```

We store too much logic in `Messenger`. I recommend creating `ENSVerifier` in `ens.go` and here we can define a single `VerifyENSNames` method.

messenger.go
```diff
@@ -916,6 +935,111 @@ func (p *postProcessor) matchMessage(message *protocol.Message, chats []*Chat) (
...
+	Error           error            `json:"error"`
+	PublicKey       *ecdsa.PublicKey `json:"-"`
+	PublicKeyString string           `json:"publicKey"`
+}
+
+func (m *Messenger) verifyEnsName(ensInfo EnsDetails, ethclient *ethclient.Client, output chan EnsResponse) {
+	publicKeyStr := ensInfo.PublicKey
+	ensName := ensInfo.Name
+	m.logger.Info("Resolving ENS name", zap.String("name", ensName), zap.String("publicKey", publicKeyStr))
+	response := EnsResponse{
```
adambabik: Any reason we have `PublicKey` in `EnsDetails` but `PublicKeyString` in `EnsResponse`?

messenger.go
```diff
@@ -916,6 +935,111 @@ func (p *postProcessor) matchMessage(message *protocol.Message, chats []*Chat) (
...
+	ch := make(chan EnsResponse)
+	response := make(map[string]EnsResponse)
+
+	ethclient, err := ethclient.DialContext(ctx, rpcEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, ensInfo := range ensDetails {
+		go m.verifyEnsName(ensInfo, ethclient, ch)
```
adambabik: Alternatively:
```suggestion
		go func() { ch <- m.verifyEnsName(ensInfo, ethclient) }()
```
so that `verifyEnsName` signature looks simpler.

messenger.go
```diff
@@ -916,6 +935,111 @@ func (p *postProcessor) matchMessage(message *protocol.Message, chats []*Chat) (
...
+
+type EnsResponse struct {
+	Name            string           `json:"name"`
+	Verified        bool             `json:"verified"`
+	Error           error            `json:"error"`
+	PublicKey       *ecdsa.PublicKey `json:"-"`
+	PublicKeyString string           `json:"publicKey"`
+}
+
+func (m *Messenger) verifyEnsName(ensInfo EnsDetails, ethclient *ethclient.Client, output chan EnsResponse) {
```
adambabik: If you want to keep this signature:
```suggestion
func (m *Messenger) verifyEnsName(ensInfo EnsDetails, ethclient *ethclient.Client, output chan<- EnsResponse) {
```

messenger.go
```diff
@@ -916,6 +935,111 @@ func (p *postProcessor) matchMessage(message *protocol.Message, chats []*Chat) (
...
+	resolver, err := ens.NewResolver(ethclient, ensName)
+	if err != nil {
+		m.logger.Error("error while creating ENS name resolver", zap.String("ensName", ensName), zap.Error(err))
+		response.Error = err
+		output <- response
+		return
+	}
+	x, y, err := resolver.PubKey()
+	if err != nil {
+		m.logger.Error("error while resolving public key from ENS name", zap.String("ensName", ensName), zap.Error(err))
```
adambabik: Recommend to handle error ones and in the most outer call possible. Maybe here:
```go
if r.Error == nil {
    // existing code
} else {
    m.logger.Error("error while resolving public key from ENS name", zap.String("ensName", ensName), zap.Error(err))
}
```

Otherwise, it looks like error is not handled in `VerifyENSNames` at all.

messenger_test.go
```diff
@@ -738,6 +738,78 @@ func (s *MessengerSuite) TestContactPersistence() {
 	s.Require().Equal(expectedContact, actualContact)
 }
 
+func (s *MessengerSuite) TestVerifyENSNames() {
+	rpcEndpoint := os.Getenv("RPC_ENDPOINT")
```
adambabik: Can we write an HTTP mock server that would act as a node exposing JSON-RPC?

messenger_test.go
```diff
@@ -738,6 +738,78 @@ func (s *MessengerSuite) TestContactPersistence() {
...
+	if rpcEndpoint == "" {
+		s.T().Skip()
+	}
+	contractAddress := "0x314159265dd8dbb310642f98f50c066173c1259b"
+	pk1 := "04325367620ae20dd878dbb39f69f02c567d789dd21af8a88623dc5b529827c2812571c380a2cd8236a2851b8843d6486481166c39debf60a5d30b9099c66213e4"
+	pk2 := "044580b6aef9ddebd88c373b43c91237dcc95a8307bc5837d11d3ad2fa5d1dc696b598e7ccc498b414ba80b86b129c48e1eb9464cc9ea26224321539f2f54024cc"
+	pk3 := "044fee950d9748606da2f77d3c51bf16134a59bde4903aa68076a45d9eefbb54182a24f0c74b381bad0525a90e78770d11559aa02f77343d172f386e3b521c277a"
+	pk4 := "not a valid pk"
+
+	ensDetails := []EnsDetails{
```
adambabik: It could be rewritten in this way:
```go
testCases := []struct{
    Name sting 
    ENSDetails ENSDetails
    Error err
    ENSResponse ENSResponse
}{
    {
        Name: "valid",
        EnsDetails: EnsDetails{Name: "pedro.stateofus.eth", PublicKey: pk1},
        EnsResponse: ...,
    },
    {
        Name: "pk not matching name",
        EnsDetails{Name: "pedro.stateofus.eth", PublicKey: pk2},
        EnsResponse: ...,
    },
    { ... },
}

for _, tc := range testCases {
    s.Run(t, func() {
        response, err := s.m.VerifyENSNames(tc.ENSDdetails, rpcEndpoint, contractAddress)
        if tc.Error != nil {
            s.EqualError(tc.Error, err.Error())
        }
        s.EqualValues(tc.ENSResponse, response)
    })
}
```

migrations/sqlite/000003_add_contacts.up.db.sql
```diff
@@ -2,10 +2,12 @@ CREATE TABLE contacts (
   id TEXT PRIMARY KEY ON CONFLICT REPLACE,
   address TEXT NOT NULL,
   name TEXT NOT NULL,
+  ens_verified BOOLEAN DEFAULT FALSE,
+  ens_verified_at INT NOT NULL DEFAULT 0,
   alias TEXT NOT NULL,
   identicon TEXT NOT NULL,
   photo TEXT NOT NULL,
-  last_updated INT NOT NULL DEFAULT 0,
+  last_updated INT DEFAULT 0,
```
adambabik: Not used?

go.mod
cammellos: Currently that throws an error:
```
# github.com/ethereum/go-ethereum/accounts/keystore
../../../../pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.4/accounts/keystore/key.go:150:49: undefined: extkeys.KeyPurpose
../../../../pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.4/accounts/keystore/key.go:158:26: extKey.ChildForPurpose undefined (type *extkeys.ExtendedKey has no field or method ChildForPurpose)
../../../../pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.4/accounts/keystore/key.go:164:26: extKey.ChildForPurpose undefined (type *extkeys.ExtendedKey has no field or method ChildForPurpose)
../../../../pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.4/accounts/keystore/keystore.go:502:40: undefined: extkeys.KeyPurposeWallet
../../../../pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.4/accounts/keystore/keystore.go:508:60: undefined: extkeys.KeyPurpose
```

Not sure what it causes it, but seems like something is breaking

messenger.go
cammellos: yes, this is much better

cammellos:
@adambabik ready to be reviewed again, the only things I have left unresolved is:
1) Geth change in go.mod, not sure but it throws an error with the previous version
2) Test for ENS

I would address both of this issue in a separate PR, as I would like to get this in and verify with status-react, if you are ok with it.
thanks!

contact.go
```diff
@@ -78,3 +87,23 @@ func existsInStringSlice(set []string, find string) bool {
 	}
 	return false
 }
+
+func buildContact(publicKey *ecdsa.PublicKey) (*Contact, error) {
+	address := strings.ToLower(crypto.PubkeyToAddress(*publicKey).Hex())
+
+	id := fmt.Sprintf("0x%s", hex.EncodeToString(crypto.FromECDSAPub(publicKey)))
```
adambabik:
```suggestion
	id := "0x" + hex.EncodeToString(crypto.FromECDSAPub(publicKey))
```

would avoid using reflection. Not important if you think the current version is more clear.

messenger.go
```diff
@@ -921,6 +910,47 @@ func Identicon(id string) (string, error) {
...
+			m.logger.Warn("Failed to resolve ens name",
+				zap.String("name", details.Name),
+				zap.String("publicKey", details.PublicKeyString),
+				zap.Error(details.Error),
+			)
+		}
+	}
+
+	if len(contacts) != 0 {
+		err = m.persistence.SetContactsENSData(contacts)
```
```suggestion
		err := m.persistence.SetContactsENSData(contacts)
```

no need to leak it outside of the block.

adambabik:
@cammellos I pushed one change. 

`github.com/status-im/status-go v0.34.0-beta.0 // indirect` is included in `go.mod` because our fork of `go-ethereum` requires it, for example:
```
github.com/ethereum/go-ethereum/accounts/keystore/key.go:	"github.com/status-im/status-go/extkeys"
```

```
replace github.com/NaySoftware/go-fcm => github.com/status-im/go-fcm v1.0.0-status
```
was needed because otherwise it couldn't find the tag.

PR #110, opened 2019-11-02
Marshal json hash

cammellos:
This is needed for status-react to correctly handle requests, as the encoding without it is different.


cammellos:
I think so, should I add those as well?

adambabik:
@cammellos no, just wanted make sure. If we haven't moved tests, we will at some point if we need them. LGTM

PR #111, opened 2019-11-04
Remove indirect status-go dependency

adambabik:
Some time ago, we introduced an indirect circular dependency (on the module level) between status-go and status-protocol-go. The artefact of that was the following entry in `go.mod`: 
```
github.com/status-im/status-go v0.34.0-beta.3 // indirect
```

The source of this circular dependency was `github.com/status-im/status-go/extkeys` which was used by our fork of go-ethereum and the fact that go-ethereum is dependency of status-protocol-go:
```
# in status-protocol-go root directory
$ go mod why -m github.com/status-im/status-go
# github.com/status-im/status-go
github.com/status-im/status-protocol-go/ens
github.com/wealdtech/go-ens/v3
github.com/ethereum/go-ethereum/accounts/abi/bind
github.com/ethereum/go-ethereum/accounts/keystore
github.com/status-im/status-go/extkeys
```
This was not a problem before because none of required go-ethereum packages was importing `github.com/status-im/status-go/extkeys` but it changed after we introduced ens support to status-protocol-go.


cammellos:
@adambabik can we merge this? I am getting into some troubles with some circular dependencies, trying to remove go-fcm (go-fcm can't be pulled in in status-go for some reason, status-protocol-go can't pull in status-go because of go-fcm, and I am not able to remove it from neither :) )

PR #113, opened 2019-11-05
Enable parsing of messages

cammellos:
This commit enables parsing of messages in status-protocol-go, status-react will conditionally use that if the message type is supported and present.

### DIFF
--- a/v1/message.go
+++ b/v1/message.go
@@ -34,8 +34,9 @@ var (
 
 // Content contains the chat ID and the actual text of a message.
 type Content struct {
-	ChatID string `json:"chat_id"`
-	Text   string `json:"text"`
+	ChatID     string `json:"chat_id"`
+	Text       string `json:"text"`
+	ResponseTo string `json:"response-to"`
 }
 
 // TimestampInMs is a timestamp in milliseconds.
--- a/v1/status_message.go
+++ b/v1/status_message.go
@@ -15,10 +15,20 @@ import (
 	statusproto "github.com/status-im/status-protocol-go/types"
 )
 
+type StatusMessageT int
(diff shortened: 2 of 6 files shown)


cammellos:
@adambabik @PombeirP ping for the review and corresponding status-go PR https://github.com/status-im/status-go/pull/1660 . Thanks!

v1/message.go
```diff
@@ -34,8 +34,9 @@ var (
 
 // Content contains the chat ID and the actual text of a message.
 type Content struct {
-	ChatID string `json:"chat_id"`
-	Text   string `json:"text"`
+	ChatID     string `json:"chat_id"`
+	Text       string `json:"text"`
+	ResponseTo string `json:"response-to"`
```
adambabik: Is it ok that `ChatID` uses snake case and `ResponseTo` uses dashes? I am fine with changing `chat_id` to `chat-id` if dashes are preferred.

v1/status_message.go
```diff
@@ -144,6 +154,13 @@ func (m *StatusMessage) HandleApplication() error {
 		return err
 	}
 	m.ParsedMessage = value
-
+	switch m.ParsedMessage.(type) {
```
adambabik: Do we have unit tests for `HandleApplication` where we could verify `MessageType` is set?

v1/message.go
cammellos: thanks, I think I will settle on a format once we have added all the fields, I have still to do a few changes on this.

v1/status_message.go
cammellos: no good point, I will add those

PR #118, opened 2019-11-06
Add name to content

cammellos:
Sometimes messages have a `name` field in the `content` map.
This is used for the ENS name.
This commit adds the parsing of that field.

### DIFF
--- a/v1/message.go
+++ b/v1/message.go
@@ -37,6 +37,7 @@ type Content struct {
 	ChatID     string `json:"chat_id"`
 	Text       string `json:"text"`
 	ResponseTo string `json:"response-to"`
+	Name       string `json:"name"` // the ENS name of the sender
 }
 
 // TimestampInMs is a timestamp in milliseconds.
--- a/v1/decoder.go
+++ b/v1/decoder.go
@@ -78,6 +78,8 @@ func statusMessageHandler(d transit.Decoder, value interface{}) (interface{}, er
 					sm.Content.Text, ok = contentVal.(string)
 				case transit.Keyword("response-to"):
 					sm.Content.ResponseTo, ok = contentVal.(string)
+				case transit.Keyword("name"):
+					sm.Content.Name, ok = contentVal.(string)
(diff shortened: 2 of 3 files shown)


cammellos:
@PombeirP @adambabik ping for a small review :) thanks!

cammellos:
And corresponding status-go pr https://github.com/status-im/status-go/pull/1664

v1/message.go
```diff
@@ -37,6 +37,8 @@ type Content struct {
 	ChatID     string `json:"chat_id"`
 	Text       string `json:"text"`
 	ResponseTo string `json:"response-to"`
+	// The ENS name of the sender
+	Name string `json:"name"`
```
adambabik:
```suggestion
	Name string `json:"name"` // the ENS name of the sender
```
so short that can be inlined :)

jakubgs: If it's the ENS name, why not call the attribute `ENSName` to make it clear?
Also, keep formatting the same, add the comment at the end of line.

cammellos: While I agree with you, currently I'd like to keep the this inline with how it is in the protocol (it is `name` there), I would revisit once this part of the codebase settles, and we can go through all the fields (I don't think this whole field should be a nested map, it's a map only for historical reasons).

jakubgs: I see, so you want to keep naming the same as in protocol spec, makes sense.

So yeah, just fix the formatting.

v1/message.go
```diff
@@ -37,6 +37,8 @@ type Content struct {
 	ChatID     string `json:"chat_id"`
 	Text       string `json:"text"`
 	ResponseTo string `json:"response-to"`
+	// The ENS name of the sender
+	Name string `json:"name"` // the ENS name of the sender
```
jakubgs: You duplicated the comment and the formatting is still wrong.

cammellos: I have applied the automatic suggestion :) will fix that

PR #119, opened 2019-11-07
Nimbus wrapper integration

PombeirP:
This PR introduces a Nimbus bridge to interface with `libnimbus.so`. It also removes the unused `MessageStore` and `ReceivedMessage` types.


bridge/nimbus/public_whisper_api.go
```diff
@@ -0,0 +1,192 @@
...
+		msg.payloadLen = C.size_t(len(req.Payload))
+		msg.payload = (*C.uchar)(C.CBytes(req.Payload))
+		defer C.free(unsafe.Pointer(msg.payload))
+		msg.paddingLen = C.size_t(len(req.Padding))
+		msg.padding = (*C.uchar)(C.CBytes(req.Padding))
+		defer C.free(unsafe.Pointer(msg.padding))
+		copyTopicToCBuffer(&msg.topic[0], req.Topic[:])
+
+		if C.nimbus_post(&msg) {
+			c <- make([]byte, 0)
```
adambabik: Why do we send an empty slice?

transport/whisper/types/whisper.go
```diff
@@ -43,12 +50,10 @@ type Whisper interface {
 	DeleteSymKey(id string) bool
 	GetSymKey(id string) ([]byte, error)
 
-	Subscribe(f Filter) (string, error)
+	Subscribe(privateKeyID, symKeyID string, pow float64, topics [][]byte) (string, error)
```
adambabik: Looks like a struct would be better. Maybe something other than `Filter` though.

bridge/nimbus/whisper.go
```diff
@@ -0,0 +1,377 @@
...
+	// return NewNimbusFilterWrapper(pFilter, id, false)
+}
+
+func (w *nimbusWhisperWrapper) Unsubscribe(id string) error {
+	return w.routineQueue.Send(func(c chan<- interface{}) {
+		idC := C.CString(id)
+		defer C.free(unsafe.Pointer(idC))
+
+		if ok := C.nimbus_unsubscribe_filter(idC); !ok {
+			c <- errors.New("filter not found")
```
adambabik: No return after sending an error? In this branch, an error and later `nil` will be sent to the channel.

bridge/nimbus/whisper.go
```diff
@@ -0,0 +1,377 @@
...
+		return "", err
+	}
+
+	retVal := w.routineQueue.Send(func(c chan<- interface{}) {
+		// Create a message store for this filter, so we can add new messages to it from the nimbus_subscribe_filter callback
+		messageList := list.New()
+		idC := C.nimbus_subscribe_filter(GetNimbusFilterFrom(f), (C.received_msg_handler)(unsafe.Pointer(C.onMessageHandler_cgo)), gopointer.Save(messageList))
+		filterID := C.GoString(idC)
+		if filterID == "" {
+			c <- errors.New("failed to subscribe to filter in Nimbus")
```
adambabik: Same here, `error` will be send and later `filterID`.

bridge/nimbus/whisper.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+func (w *nimbusWhisperWrapper) AddSymKeyDirect(key []byte) (string, error) {
+	retVal := w.routineQueue.Send(func(c chan<- interface{}) {
+		keyC := C.CBytes(key)
+		defer C.free(unsafe.Pointer(keyC))
+
+		idC := C.nimbus_add_symkey((*C.uchar)(unsafe.Pointer(keyC)))
+		if idC == nil {
+			c <- errors.New("failed to add symkey to Nimbus")
```
adambabik: And here.

bridge/nimbus/whisper.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+func (w *nimbusWhisperWrapper) AddSymKeyFromPassword(password string) (string, error) {
+	retVal := w.routineQueue.Send(func(c chan<- interface{}) {
+		passwordC := C.CString(password)
+		defer C.free(unsafe.Pointer(passwordC))
+
+		idC := C.nimbus_add_symkey_from_password(passwordC)
+		if idC == nil {
+			c <- errors.New("failed to add symkey to Nimbus")
```
adambabik: And here.

bridge/nimbus/routine_queue.go
```diff
@@ -0,0 +1,54 @@
...
+		return
+	}
+}
+
+// Send executes the passed function. This method can be called safely from a
+// goroutine in order to execute a Nimbus function. It is important to note that the
+// passed function won't be executed immediately, instead it will be added to
+// the user events queue.
+func (q *RoutineQueue) Send(f func(chan<- interface{})) interface{} {
+	ev := event{f: f, done: make(chan interface{}, 1)}
```
adambabik: It's fine but closing `done` channel might be useful to avoid bugs like sending more than one message to it.

bridge/nimbus/public_whisper_api.go
PombeirP: This is temporary code until Nimbus starts returning envelope hashes on success. I'll ping @kdeme to see if we can get that in Nimbus, this kind of slipped through the cracks as it is not required right now.

transport/whisper/types/whisper.go
PombeirP: Done. Added `SubscriptionOptions`.

bridge/nimbus/whisper.go
PombeirP: Oops, good catch! I wish Go had some kind of language mechanism to make code like this less repetitive (such as aspect-oriented programming support or AST macros).

bridge/nimbus/routine_queue.go
PombeirP: Good point, adding `defer close(ev.done)`

bridge/nimbus/public_whisper_api.go
PombeirP: Kim said he'll implement this in Nimbus in the afternoon, and then I'll be able to make the necessary changes on the Go side.

PombeirP: I've discussed with Adam and we agreed to leave this for a subsequent PR, since the envelope hash isn't being used anywhere for the Nimbus implementation.

PR #122, opened 2019-11-13
Add markdown support

cammellos:
This commits introduce parsing of basic markdown in status-go.

To do this we use a stripped down version of https://github.com/gomarkdown/markdown , customized to support some extra use cases and JSON encoding. The code for the library is still a bit rough and I need to go around fixing tests etc, but keen to get the changes merged in status-react so I don't have to rebase constantly :) . https://github.com/status-im/markdown

We expose an endpoint to parse content for now until the sending of the message is fully in status-protocol-go.

Corresponding status-go/status-react PR: https://github.com/status-im/status-go/pull/1674 https://github.com/status-im/status-react/pull/9409


messenger.go
```diff
@@ -811,6 +811,17 @@ func (m *Messenger) RetrieveRawAll() (map[transport.Filter][]*protocol.StatusMes
 				continue
 			}
 
+			for _, msg := range statusMessages {
+				if msg.ParsedMessage != nil {
+					if _, ok := msg.ParsedMessage.(protocol.Message); ok {
```
```suggestion
					if textMessage, ok := msg.ParsedMessage.(protocol.Message); ok {
```

messenger.go
```diff
@@ -811,6 +811,17 @@ func (m *Messenger) RetrieveRawAll() (map[transport.Filter][]*protocol.StatusMes
...
+						textMessage := msg.ParsedMessage.(protocol.Message)
```
```suggestion
```

v1/content.go
```diff
@@ -0,0 +1,54 @@
...
+	Name       string   `json:"name"` // the ENS name of the sender
+	ParsedText ast.Node `json:"parsedText"`
+	LineCount  uint     `json:"lineCount"`
+	RTL        bool     `json:"rtl"`
+}
+
+// Check if the first character is Hebrew or Arabic or the RTL character
+func isRTL(s string) bool {
+
+	var first rune
```
adambabik: `utf8.DecodeRuneInString` maybe instead of the loop?

v1/content.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	return unicode.Is(unicode.Hebrew, first) ||
+		unicode.Is(unicode.Arabic, first) ||
+		// RTL character
+		first == '\u200f'
+}
+
+// calculateLineCount counts the number of newlines characters in the message
+func calculateLineCount(text string) uint {
```
adambabik: `strings.Count`?

cammellos:
@adambabik addressed feedback, thanks!

v1/content.go
```diff
@@ -0,0 +1,39 @@
...
+	ResponseTo string   `json:"response-to"`
+	Name       string   `json:"name"` // the ENS name of the sender
+	ParsedText ast.Node `json:"parsedText"`
+	LineCount  uint     `json:"lineCount"`
+	RTL        bool     `json:"rtl"`
+}
+
+// Check if the first character is Hebrew or Arabic or the RTL character
+func isRTL(s string) bool {
+
```
adambabik:
```suggestion
```

v1/content.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// Content contains the chat ID and the actual text of a message.
+type Content struct {
+	ChatID     string   `json:"chat_id"`
+	Text       string   `json:"text"`
+	ResponseTo string   `json:"response-to"`
+	Name       string   `json:"name"` // the ENS name of the sender
+	ParsedText ast.Node `json:"parsedText"`
+	LineCount  uint     `json:"lineCount"`
```
adambabik: Any reason to use `uint` instead of `int`?

cammellos: just because it can't be negative, should I change it to int instead?

cammellos:
@adambabik no, we should probably document it somewhere, currently we support (as well as code blocks):
```
*italic* or _italic_
**bold** or __bold__
http://link.com or https://link.com
`code`
> quote
#status-tag
```

v1/content.go
adambabik: Can it be realistically set to a negative number? `len(slice)` also can't be negative but it uses `int`. If you are fine with type casting and `unint` type carries some useful information that is missing when using `int`, let's keep it.

cammellos: no, will never be negative, i just went instinctively for uint, what s more customary in this case? happy to go for the one feels more natural ( seens like int is)

adambabik: More natural would be `int` because its name is `LineCount` so we know it's only positive. If it was a continuous counter like a metric counter, `int64` would be better. If a counter was serialized to bytes and send through the network, then I would go with `int32` or `int64` or even unsigned ints.

But I actually don't know the reason why `len()` returns `int`. The only thing that comes to my mind is convenience. Dealing with unsigned ints is harder and can result in odd values in the case of overflow.

